PR #1, opened 2018-06-21
Set up test suite with dummy test

zuren:
- Using XUnit as it's Hackney standard, and seems more modern than NUnit.
- Adds an example "Hello world" test.

### DIFF
--- a/LBHTenancyAPI/Test/Foo/GreeterTest.cs
+++ b/LBHTenancyAPI/Test/Foo/GreeterTest.cs
@@ -1,16 +1,15 @@
-﻿using NUnit.Framework;
+﻿using Xunit;
 using LBHTenancyAPI.Foo;
 
 namespace LBHTenancyAPI.Test.Foo
 {
-    [TestFixture]
     public class GreeterTest
     {
-        [Test]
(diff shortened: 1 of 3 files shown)


LBHTenancyAPI/Test/Foo/GreeterTest.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Xunit;
+using LBHTenancyAPI.Foo;
+
+namespace LBHTenancyAPI.Test.Foo
+{
+    public class GreeterTest
+    {
+        [Fact]
```
StevenLeighton21: These decorators 👍

LBHTenancyAPI/Test/Foo/GreeterTest.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Xunit;
+using LBHTenancyAPI.Foo;
+
+namespace LBHTenancyAPI.Test.Foo
+{
+    public class GreeterTest
```
VladyslavAtamanyuk: LGTM

PR #7, opened 2018-07-03
Dockerised sqlserver

StevenLeighton21:
- Add a dockerised mssql server
- Add this to the docker startup
- Add a very basic sql server setup script - creates a Test table
- Add a test that demonstrates very basic AACT flow for interacting with the server during tests

TODO:
Add a script that will start up the stub locally or check it is running before running tests (or add something to the tests?)
Add a linter


LBHTenancyAPI/Dockerfile
```diff
@@ -9,10 +9,10 @@ COPY LBHTenancyAPI/LBHTenancyAPI.csproj LBHTenancyAPI/
 RUN dotnet restore -nowarn:msb3202,nu1503
 COPY . .
 WORKDIR /src/LBHTenancyAPI
-RUN dotnet build -c Release -o /app
+RUN dotnet build LBHTenancyAPI.csproj -c Release -o /app
 
 FROM build AS publish
-RUN dotnet publish -c Release -o /app
+RUN dotnet publish LBHTenancyAPI.csproj -c Release -o /app
```
zuren: Will this cause issues in CI?

LBHTenancyAPI/LBHTenancyAPI.csproj
```diff
@@ -1,21 +1,16 @@
...
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
   </ItemGroup>
-
   <ItemGroup>
     <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="2.0.0" />
   </ItemGroup>
-
-</Project>
+</Project>
```
zuren: Did we manually modify this file or was it the IDE? Should we be conscious of unnecessary changes as this could flip flop back and forth if they generate differently?

LBHTenancyAPI/LBHTenancyAPI.sln
```diff
@@ -0,0 +1,17 @@
...
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{BFBB2C9F-E0A5-4249-A219-1ACF6DCC3B43}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{BFBB2C9F-E0A5-4249-A219-1ACF6DCC3B43}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{BFBB2C9F-E0A5-4249-A219-1ACF6DCC3B43}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{BFBB2C9F-E0A5-4249-A219-1ACF6DCC3B43}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+EndGlobal
```
zuren: Why do we have two LBHTenancyAPI solution files now?

LBHTenancyAPI/Models/Test.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace LBHTenancyAPI.Models
+{
+    public class Test
+    {
+        public int ID { get; set; }
+        public string Name { get; set; }
+    }
+}
```
zuren: Should this be in LBHTenancyAPI as it's exclusively a test class?

LBHTenancyAPI/docker-compose.dcproj
```diff
@@ -0,0 +1,11 @@
...
+<Project ToolsVersion="15.0" Sdk="Microsoft.Docker.Sdk" DefaultTargets="Build">
+  <PropertyGroup Label="Globals">
+    <ProjectVersion>2.1</ProjectVersion>
+    <DockerTargetOS>Linux</DockerTargetOS>
+    <ProjectGuid>{C5027587-95A1-4BFB-B7A5-5587E26E804C}</ProjectGuid>
+    <DockerLaunchBrowser>True</DockerLaunchBrowser>
+    <DockerServiceUrl>http://localhost:{ServicePort}</DockerServiceUrl>
+    <DockerServiceName>lbhtenancyapi</DockerServiceName>
+  </PropertyGroup>
+</Project>
```
zuren: Didn't we already have a dcproj file too?

LBHTenancyAPITest/Test/SqlGateway/SqlGatewayTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        
+        /*
+         * Setup test context in constructor
+         */
+        public SqlGatewayTest(DatabaseFixture fixture)
+        {
+            db = fixture;
+            db.Db.Open();
+            db.Db.Query("Insert into Test values (1, 'banana')");
+
```
zuren: Rogue newline.

LBHTenancyAPI/LBHTenancyAPI.csproj
StevenLeighton21: IDE - It's awful and I don't know of a way around it

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
```diff
@@ -0,0 +1,30 @@
...
+            builder.Password = "Rooty-Tooty";
+            Db = new SqlConnection(builder.ConnectionString);
+        }
+
+        /*
+        * Called when test runner releases fixture
+        */
+        public void Dispose()
+        {
+            Db.Query("DELETE FROM Test");
```
zuren: Is it possible to delete from all tables instead?

LBHTenancyAPI/LBHTenancyAPI.sln
StevenLeighton21: We don't, we have two GUUID references for the same solution - this is so you IDE settings persist (probably due to us using Rider and VS mixed)

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public SqlConnection Db { get; private set; }
+        public DatabaseFixture()
+        {
+            SqlConnectionStringBuilder builder =  new SqlConnectionStringBuilder();  
+            builder["Data Source"] = "(local)";
+            builder["integrated Security"] = false;
+            builder["Initial Catalog"] = "StubUH";
+            builder.UserID = "sa";
+            builder.Password = "Rooty-Tooty";
```
zuren: Should we begin splitting this out into environment config, or is that a job for another PR?

LBHTenancyAPI/Models/Test.cs
StevenLeighton21: I imagine it will go away at some point or be renamed to a more useful model....I put it here to demonstrate where Dapper models will live more than anything

StubUniversalHousing/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# mssql-server-linux
+# Maintainers: Microsoft Corporation (LuisBosquez and twright-msft on GitHub)
+# GitRepo: https://github.com/Microsoft/mssql-docker
+
+# Base OS layer: Latest Ubuntu LTS.
+#FROM ubuntu:16.04
+FROM microsoft/mssql-server-linux
```
zuren: Must clean up the comments in this file.

LBHTenancyAPI/docker-compose.dcproj
StevenLeighton21: Welcome to the wonderful world of IDE development

StubUniversalHousing/Program.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            BuildWebHost(args).Run();
+        }
+
+        public static IWebHost BuildWebHost(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+                .UseStartup<Startup>()
+                .Build();
+    }
+}
```
zuren: Do we need this program file? Probably not?

StubUniversalHousing/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
...
+    "StubUniversalHousing": {
+      "commandName": "Project",
+      "launchBrowser": true,
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      },
+      "applicationUrl": "http://localhost:60444/"
+    }
+  }
+}
```
zuren: Same here

StubUniversalHousing/Startup.cs
```diff
@@ -0,0 +1,34 @@
...
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.Run(async (context) =>
+            {
+                await context.Response.WriteAsync("Hello World!");
+            });
+        }
+    }
+}
```
zuren: And here

LBHTenancyAPITest/Test/SqlGateway/SqlGatewayTest.cs
LBHRShetty: I still don't understand why did we use LBHTenancyAPI.Models; inside our LBHTenancyAPITest.Test.SqlGateway namespace

LBHTenancyAPI/Models/Test.cs
VladyslavAtamanyuk: It has it's own folder and namespace now.

LBHTenancyAPITest/Test/SqlGateway/SqlGatewayTest.cs
LBHRShetty: Explain Assert.Single(entries);

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: Not at this point - I'd suggest we eventually have a private function in the fixture that drops tables individually and you add tables to it as you add them - also the AAAT flow will naturally mean dispose mirrors arrange.

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: I am not sure we even need to, this database is test only and basically meaningless. If you have access to this code you'll be able to see the schema and create your own stub locally, but not access mine.

StubUniversalHousing/Dockerfile
StevenLeighton21: 👍

StubUniversalHousing/Program.cs
LBHRShetty: We don't need this at the moment but we will need to write startup class later

LBHTenancyAPI/LBHTenancyAPI.sln
zuren: Seems bad, uncomfortable with this.

LBHTenancyAPITest/Test/SqlGateway/SqlGatewayTest.cs
StevenLeighton21: Asserts the collection contains a single item - looks nicer than assert(1, entries.count)

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: Why do it individually?

LBHTenancyAPITest/Test/SqlGateway/SqlGatewayTest.cs
VladyslavAtamanyuk: Assert.Single(something) will check that something will occur once.

StevenLeighton21: It seemed to work with the namespace in the class but not at the top of the file. Our minimal linting doesn't seem to care so I'm happy to leave it in place.

LBHTenancyAPI/LBHTenancyAPI.sln
StevenLeighton21: S A M E

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: Control, readability - you see what tables are there in create, you see what tables are deleted in dispose. I'm personally uncomfortable with blanket drop tables statements.

docker-compose.override.yml
```diff
@@ -1,8 +1,15 @@
...
+    environment:
+      - ASPNETCORE_ENVIRONMENT=Development
+    ports:
+      - "80"
+
+  stubuniversalhousing:
+    environment:
+      - ASPNETCORE_ENVIRONMENT=Development
+    ports:
+      - "80"
```
VladyslavAtamanyuk: Should these sit on different ports?

zuren: Good point, what's VS up to here? ^

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: I see your point, I still quite like the rspec pattern of wiping out the db between every test run though. It ensures there's no way to keep state around between tests, if you forget to delete the table's entries, which can cause confusing failures IMO. Also reduces the size of each test which is nice, they're fairly verbose.

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: I expect in actual production code, we'll be constructing the same code from env config, so may be simpler to reuse that once we do. LGTM for now.

docker-compose.override.yml
StevenLeighton21: I don't think it matters in the override since the stub will never care about port 80, though I guess it's best to avoid potential port binding clashes

LBHTenancyAPI/LBHTenancyAPI.sln
StevenLeighton21: Misunderstood this. Skills. I've seen duplicate GUUID entries in IDE cfg files before. What happened here was that we began with just the API project and eventually added more in an umbrella project and left a vestigial .sln file.

LBHTenancyAPI/Dockerfile
StevenLeighton21: Don't believe so - it simply specifies which project to build which is now necessary as we have projects we may not wish to build in the solution

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: Dispose is done in the fixture - so I envision adding table structure in the setup.sql file, and adding content to test against in the test setup, and then all tables get wiped when the fixture is disposed. I am not sure this is the best approach (does the fixture persist between tests? I believe it's a new instance for each test class). The reason I want to delete contents in the dispose is so that everything is deleted in one place - everything is created in one place outside the tests by necessity, so I'd like deletion to mirror that so it can be easily compared.

LBHTenancyAPITest/Test/SqlGateway/SqlGatewayTest.cs
```diff
@@ -0,0 +1,35 @@
...
+            db = fixture;
+            db.Db.Open();
+            db.Db.Query("Insert into Test values (1, 'banana')");
+        }
+
+        [Fact]
+        public void GetAllTenanciesTest()
+        {
+            const string sql = "SELECT * FROM Test";
+            List<Test> entries = db.Db.Query<LBHTenancyAPI.Models.Test>(sql).ToList();
```
zuren: Could be `<Test>` now I think?

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
```diff
@@ -0,0 +1,39 @@
...
+        public DatabaseFixture()
+        {
+            SqlConnectionStringBuilder builder =  new SqlConnectionStringBuilder();  
+       
+            builder["Data Source"] = "(local)";
+            
+            if (Environment.GetEnvironmentVariable("CI_TEST") == "True")
+            {
+                builder["Data Source"] = "tcp:stubuniversalhousing";
+            }
```
zuren: Why not pass the data source in as an environment variable instead of switching?

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
```diff
@@ -0,0 +1,39 @@
...
+            builder["integrated Security"] = false;
+            builder["Initial Catalog"] = "StubUH";
+            builder.UserID = "sa";
+            builder.Password = "Rooty-Tooty";
+            Db = new SqlConnection(builder.ConnectionString);
+        }
+
+        /*
+        * Called when test runner releases fixture
+        */
```
zuren: Not sure this comment is necessary? Same with the "Setup test context in constructor" comment in SqlGatewayTest.cs.

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: This way it doesn't matter if the environment isn't set when developing locally, and facilitates IDE development and running as well as CLI style development. It also means you only need to worry about setting it for CI. I probably should just look for it to be set at all rather than to a specific value,

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: Doesn't seem at odds with wiping out all of them?

README.md
```diff
@@ -0,0 +1,9 @@
+## Setup
+
+## Running the tests
+Run `docker-compose run --rm stubuniversalhousing` to bring up the stub data source. Tests run against this in order to simulate the legacy data source.
+
+You can now run the tests using the test explorer in your IDE (we've used Rider and Visual Studio CE on Mac)
+
+TODO: Add xunit console runner to run tests from cli
```
VladyslavAtamanyuk: I think this can be taken out now, right?

sql_endpoints.md
```diff
@@ -0,0 +1,117 @@
...
+    "payment_method": {
+      "code": "🚫",
+      "display_name": "🚫"
+    },
+    "delta": "🚫",
+    "final_balance": "🚫"
+  }],
+  "number_of_transactions": "🚫",
+}
+```
```
zuren: Can delete this, did you branch from sql-endpoints or something?

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: I put them in to clarify _when_ arrange and teardown happen, as it is not immediately obvious if you're not familiar with our setup. I feel it alleviates one of the concerns given around how and where we're doing teardown in terms of possible maintenance issues.

zuren:
You mean AAAT flow?

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: Isn't this just typical xunit setup though, not particular to our app? Feels like something people will be generally aware of, may be superfluous. Your call though.

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: Yeah I guess it doesn't! :)

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: Yeah probably. Have you confirmed it doesn't fall over when run without?

StevenLeighton21: I don't have it set on my local machine, it's working fine in IDE dev. It will fall over if you do set it locally....
I'm sure there's a better way to handle it. I'd rather get this merged so we can develop and accept that there's a future task to improve this.

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
StevenLeighton21: We weren't aware but we certainly could appeal to the maintainers being more familiar with xunit that we are.

LBHTenancyAPI/Models/Test.cs
LBHRShetty: Great to have model in place

LBHTenancyAPITest/Test/StubUniversalHousingFixture.cs
zuren: Cool 👍

PR #9, opened 2018-07-06
Add editor config, adjust code style rules, apply code style

StevenLeighton21:
- Added .editorconfig
 - Tweaked some of the code style settings in Rider (saved to shared layer)
 - Applied new settings to code
 - I've added specific exceptions in our test model - the rules make sense for other classes so did not want to globally disable.


LBHTenancyAPI/Controllers/ValuesController.cs
```diff
@@ -15,7 +11,7 @@ public class ValuesController : Controller
         public string Get()
         {
             Greeter greeter = new Greeter();
```
VladyslavAtamanyuk: the var greeter should be capital G right?

LBHTenancyAPI/Models/Test.cs
```diff
@@ -1,8 +1,12 @@
-﻿namespace LBHTenancyAPI.Models
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace LBHTenancyAPI.Models
 {
+    [SuppressMessage("ReSharper", "ClassNeverInstantiated.Global")]
```
VladyslavAtamanyuk: Are we using reSharper utilities? Not everyone has reSharper is this still ok?

LBHTenancyAPITest/Test/Foo/GreeterTest.cs
```diff
@@ -9,7 +9,7 @@ public class GreeterTest
         public void SaysHello()
         {
             Greeter greeter = new Greeter();
-            Assert.Equal("hello world", greeter.SayHello());
+            Assert.Equal("hello world", Greeter.SayHello());
```
VladyslavAtamanyuk: Capital G on the constructed Greeter object?

LBHTenancyAPI/Models/Test.cs
StevenLeighton21: If you're using VS you _should_ use resharper as I understand it.

There's a separate discussion around using things that are compatible with people's personal IDE choices - in order to maintain code style you should use something linter compatible

PR #10, opened 2018-07-09
Create UHTenancies gateway that retrieves tenancy list details

StevenLeighton21:
- Create gateway to retrieve tenancy list items
- Added .env support for configuring UH connection
- Moved from models to domain objects
- Stub has had `contacts`, `tenagree`, `arag` and `araction` tables added (not full schema, just columns we retrieve data from)


.circleci/config.yml_old
```diff
@@ -0,0 +1,99 @@
+version: 2
```
zuren: No thanks

.gitignore
```diff
@@ -39,3 +39,5 @@ Thumbs.db
 
 # dotCover
 *.dotCover
+
+.env*
```
zuren: This `*` isn't including `.env` itself. Should probably have:

```sh
.env
.env.*
```

LBHTenancyAPI/Program.cs
```diff
@@ -13,7 +13,7 @@ namespace LBHTenancyAPI
     public class Program
     {
         public static void Main(string[] args)
-        {
+        {
```
zuren: Why?

LBHTenancyAPI/Startup.cs
```diff
@@ -29,6 +30,9 @@ public void ConfigureServices(IServiceCollection services)
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
+            string dotenv = Path.GetRelativePath(Directory.GetCurrentDirectory(), "../../../.env");
```
zuren: Is this a reliable way to read configuration in dev/staging/production? Have we tested it yet?

zuren:
There are some major holes in this implementation atm, but recommend we change them in a later PR:

- No test case (or implementation for) current agreement.
- No test case (or implementation for) latest arrears action event.
- No test case (or implementation for) short address - instead showing full address.
- No guarantee that test database schema lines up with the real UH schema.

zuren:
Also, as an optimisation, we could skip the SQL query altogether when `tenancyRefs` is an empty list.

LBHTenancyAPI/Startup.cs
StevenLeighton21: Works now, by defaulting to the option passed in the circle cfg.yml. Need a more concrete solution for production IMO

zuren: Lets leave it for now, it's only really relevant in dev so we'll likely change it elsewhere.

PR #13, opened 2018-07-10
Added only latest arrears action to list item

StevenLeighton21:
- Return only the most recent arrears action with list item
- Refactored tests slightly


LBHTenancyAPI/Domain/TenancyListItem.cs
```diff
@@ -13,5 +13,60 @@ public struct TenancyListItem
         public string PrimaryContactName { get; set; }
         public string PrimaryContactShortAddress { get; set; }
         public string PrimaryContactPostcode { get; set; }
+
+        public TenancyListItem mergeWith(TenancyListItem t)
+        {
+            if (t.TenancyRef != null && TenancyRef == null)
```
VladyslavAtamanyuk: Just wanted to clarify, are we assuming that the TenancyListItem TenancyRef will always have a value and never be null?

LBHRShetty: cool.

StevenLeighton21: We're assuming self is truthy and that t will have some values to replace null values in self, so we only grab those where that is the case :)

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -83,20 +94,24 @@ private TenancyListItem CreateRandomTenancyListItem()
...
+            InsertTenancyAttributes(tenancy);
+
+            return tenancy;
+        }
+
         private void InsertTenancyAttributes(TenancyListItem tenancyAttributes)
         {
             string commandText =
-                "INSERT INTO araction (tag_ref, action_code, action_date) VALUES (@tenancyRef, @lastActionType, @lastActionTime);" +
+
```
VladyslavAtamanyuk: Wild empty line

LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
```diff
@@ -19,27 +20,49 @@ public UhTenanciesGateway(string connectionString)
...
+
+                result.TenancyRef = row[0].TenancyRef;
+                result.LastActionCode = row[0].LastActionCode;
+                result.LastActionDate = row[0].LastActionDate;
+                result = tenancyListItem.mergeWith(result);
+
+                results.Add(result);
+            }
+
+            return results;
```
zuren: Why is this necessary? Feel like we ought to be able to:
 - Order by arag.start_date and araction.action_date to make the top result contain the latest of each.
- Make rows distinct on tag_ref to ensure we only get one result per tenancy.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -57,6 +52,22 @@ public void WhenGivenTenancyRef_GetTenanciesByRefs_ShouldReturnTheLatestAgreemen
...
         }
 
+        [Fact]
+        public void WhenGivenTenancyRef_GetTenanciesByRefs_ShouldReturnTheLatestArrearsAction()
+        {
+            TenancyListItem expectedTenancy = InsertRandomisedTenancyListItem();
+
+            DateTime latestActionDate = expectedTenancy.LastActionDate.AddDays(1);
+            InsertArrearsActions(expectedTenancy.TenancyRef, "ABC",
+                expectedTenancy.LastActionDate.Subtract(DAY_IN_TIMESPAN));
```
zuren: The `Subtract` isn't necessary, right?

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -125,5 +141,22 @@ private void InsertAgreement(string tenancyRef, string status, DateTime startDat
 
             command.ExecuteNonQuery();
         }
+
+
```
VladyslavAtamanyuk: Wild empty line here

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -57,6 +52,22 @@ public void WhenGivenTenancyRef_GetTenanciesByRefs_ShouldReturnTheLatestAgreemen
...
+            TenancyListItem expectedTenancy = InsertRandomisedTenancyListItem();
+
+            DateTime latestActionDate = expectedTenancy.LastActionDate.AddDays(1);
+            InsertArrearsActions(expectedTenancy.TenancyRef, "ABC",
+                expectedTenancy.LastActionDate.Subtract(DAY_IN_TIMESPAN));
+            InsertArrearsActions(expectedTenancy.TenancyRef, "XYZ", latestActionDate);
+
+            var tenancies = GetTenanciesByRef(new List<string> {expectedTenancy.TenancyRef});
+
+            Assert.Equal(tenancies[0].LastActionDate, latestActionDate);
```
zuren: Wonder if we should have a guarantee that this is using date to sort, instead of insertion order? At the moment, it could be either.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -125,5 +141,22 @@ private void InsertAgreement(string tenancyRef, string status, DateTime startDat
 
             command.ExecuteNonQuery();
         }
+
```
VladyslavAtamanyuk: Another Wild Line Here

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
StevenLeighton21: Seems reasonable to have a randomised date then one entry before it and one after

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
StevenLeighton21: No, it's adding the date we should get last, where the query unsorted would return the rows FIFO, no?

PR #14, opened 2018-07-10
Return the short address (up to first newline) in address

StevenLeighton21:
- Future PR - 
Compare the schema with our test schema, ensure accuracy
Use a proper format for the current balance
Treat every value in the DB as nullable (pretty large refactor/ addition of tests)

### DIFF
--- a/LBHTenancyAPI/Domain/TenancyListItem.cs
+++ b/LBHTenancyAPI/Domain/TenancyListItem.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 
 namespace LBHTenancyAPI.Domain
 {
@@ -11,7 +12,15 @@ namespace LBHTenancyAPI.Domain
         public string ArrearsAgreementStatus { get; set; }
         public DateTime ArrearsAgreementStartDate { get; set; }
         public string PrimaryContactName { get; set; }
-        public string PrimaryContactShortAddress { get; set; }
+
+        private string primaryContactShortAddress;
+        public string PrimaryContactShortAddress
+        {
+            get => primaryContactShortAddress;
+
+            set => primaryContactShortAddress = !string.IsNullOrEmpty(value) ? value.Split("\n").First() : null;
+        }
+
         public string PrimaryContactPostcode { get; set; }
     }
 }
--- a/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
+++ b/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
@@ -114,6 +114,7 @@ namespace LBHTenancyAPITest.Test.Gateways
             InsertTenancyAttributes(expectedTenancy);
 
             var tenancies = GetTenanciesByRef(new List<string> {expectedTenancy.TenancyRef});
+
             Assert.Equal(expectedTenancy.PrimaryContactShortAddress, tenancies[0].PrimaryContactShortAddress);
         }
 
@@ -139,7 +140,6 @@ namespace LBHTenancyAPITest.Test.Gateways
                 PrimaryContactName = random.Word(),
                 PrimaryContactShortAddress = $"{random.Words()}\n{random.Words()}\n{random.Words()}\n{random.Words()}",
                 PrimaryContactPostcode=  random.Word()
-
             };
         }
 
--- a/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
+++ b/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
@@ -94,6 +94,29 @@ namespace LBHTenancyAPITest.Test.Gateways
             Assert.Equal(secondTenancyLatestAgreementStartDate, receivedSecond.First().ArrearsAgreementStartDate);
         }
 
+        [Fact]
+        public void WhenGivenATenancyRef_GetTenanciesByRefs_ShouldReturnOnlyTheShortAddress()
+        {
+            TenancyListItem expectedTenancy = InsertRandomisedTenancyListItem();
+            string longAddress = expectedTenancy.PrimaryContactShortAddress;
+            string actualShortAddressExpected = longAddress.Split("\n")[0];
+
+            var tenancies = GetTenanciesByRef(new List<string> {expectedTenancy.TenancyRef});
+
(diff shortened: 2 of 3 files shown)


LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -94,6 +94,29 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
...
 
+        [Fact]
+        public void WhenGivenATenancyRef_GetTenanciesByRefs_ShouldReturnOnlyTheShortAddress()
+        {
+            TenancyListItem expectedTenancy = InsertRandomisedTenancyListItem();
+            string longAddress = expectedTenancy.PrimaryContactShortAddress;
+            string actualShortAddressExpected = longAddress.Split("\n")[0];
+
+            var tenancies = GetTenanciesByRef(new List<string> {expectedTenancy.TenancyRef});
+
```
VladyslavAtamanyuk: Wild empty line.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -114,7 +137,7 @@ private TenancyListItem CreateRandomTenancyListItem()
                 ArrearsAgreementStatus = random.Word(),
                 ArrearsAgreementStartDate = new DateTime(random.Int(1900, 1999), random.Int(1, 12), random.Int(1, 28), 9, 30, 0),
                 PrimaryContactName = random.Word(),
-                PrimaryContactShortAddress = random.Words(),
+                PrimaryContactShortAddress = $"{random.Words()}\n{random.Words()}\n{random.Words()}\n{random.Words()}",
                 PrimaryContactPostcode=  random.Word()
```
VladyslavAtamanyuk: Wild white space

LBHTenancyAPI/Domain/TenancyListItem.cs
```diff
@@ -11,7 +12,15 @@ public struct TenancyListItem
...
         public DateTime ArrearsAgreementStartDate { get; set; }
         public string PrimaryContactName { get; set; }
-        public string PrimaryContactShortAddress { get; set; }
+
+        private string primaryContactShortAddress;
+        public string PrimaryContactShortAddress
+        {
+            get => primaryContactShortAddress;
+
+            set => primaryContactShortAddress = !string.IsNullOrEmpty(value) ? value.Split("\n").First() : null;
```
zuren: Feel slightly uncomfortable that this logic is in the domain object rather than the class providing the ShortAddress, feels like more moving parts so more room for confusion - why did you choose to put it here?

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -94,6 +94,30 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
             Assert.Equal(secondTenancyLatestAgreementStartDate, receivedSecond.First().ArrearsAgreementStartDate);
         }
 
+        [Fact]
+        public void WhenGivenATenancyRef_GetTenanciesByRefs_ShouldReturnOnlyTheShortAddress()
+        {
+            TenancyListItem expectedTenancy = InsertRandomisedTenancyListItem();
+            string longAddress = expectedTenancy.PrimaryContactShortAddress;
+            string actualShortAddressExpected = longAddress.Split("\n")[0];
```
zuren: `var`?

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -94,6 +94,30 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
...
+
+            var tenancies = GetTenanciesByRef(new List<string> {expectedTenancy.TenancyRef});
+
+            Assert.Equal(actualShortAddressExpected, tenancies[0].PrimaryContactShortAddress);
```
zuren: Given my comment above, believe this code isn't actually testing the UhTenanciesGateway at all. The address will have already been snipped to a single line before it's saved to the database, given it's set on a TenancyListItem object.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -94,6 +94,30 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
...
+        public void WhenGivenATenancyRefWithNoAddress_GetTenanciesByRefs_ShouldReturnNull()
+        {
+            TenancyListItem expectedTenancy = CreateRandomTenancyListItem();
+            expectedTenancy.PrimaryContactShortAddress = null;
+            InsertTenancyAttributes(expectedTenancy);
+
+            var tenancies = GetTenanciesByRef(new List<string> {expectedTenancy.TenancyRef});
+
+            Assert.Equal(expectedTenancy.PrimaryContactShortAddress, tenancies[0].PrimaryContactShortAddress);
+        }
```
zuren: How about PrimaryContactName and PrimaryContactPostcode? Know these will be missing from some cases as well.

LBHTenancyAPI/Domain/TenancyListItem.cs
StevenLeighton21: Dapper is mapping the results of the query directly to this object - it felt like an elegant solution to override the setter and ensure that libraries auto hitting the setter get caught.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
StevenLeighton21: Technically, it does test the implementation....

But great catch!

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
StevenLeighton21: Future PR - as I understand it this should have an API wide refactor to treat all fields as nullable

PR #15, opened 2018-07-11
Refactoring gateways

StevenLeighton21:
- Use schema types
- Get tests passing
- Properly retrieve the lates action and agreement (requires further refactor)
- Use a decimal for balance, round to 2 places on retrieval (C# has built in rounding for currency - suggest we look at this, though we're constrained by whatever UH is currently doing)

### DIFF
--- a/LBHTenancyAPI/Domain/TenancyListItem.cs
+++ b/LBHTenancyAPI/Domain/TenancyListItem.cs
@@ -8,7 +8,7 @@ namespace LBHTenancyAPI.Domain
         public string TenancyRef { get; set; }
         public string LastActionCode { get; set; }
         public DateTime LastActionDate { get; set; }
-        public double CurrentBalance { get; set; }
+        public decimal CurrentBalance { get; set; }
         public string ArrearsAgreementStatus { get; set; }
         public DateTime ArrearsAgreementStartDate { get; set; }
         public string PrimaryContactName { get; set; }
--- a/LBHTenancyAPI/Domain/TenancyListItem.cs
+++ b/LBHTenancyAPI/Domain/TenancyListItem.cs
@@ -8,17 +8,27 @@ namespace LBHTenancyAPI.Domain
         public string TenancyRef { get; set; }
         public string LastActionCode { get; set; }
         public DateTime LastActionDate { get; set; }
-        public decimal CurrentBalance { get; set; }
+
+        private decimal currentBalance;
+
+        public Decimal CurrentBalance
+        {
+            get => currentBalance;
+
+            set => currentBalance = decimal.Round(value, 2, MidpointRounding.AwayFromZero);
+        }
+
         public string ArrearsAgreementStatus { get; set; }
         public DateTime ArrearsAgreementStartDate { get; set; }
         public string PrimaryContactName { get; set; }
 
         private string primaryContactShortAddress;
+
         public string PrimaryContactShortAddress
         {
             get => primaryContactShortAddress;
 
-            set => primaryContactShortAddress = !string.IsNullOrEmpty(value) ? value.Split("\n").First() : null;
+            set => primaryContactShortAddress = string.IsNullOrWhiteSpace(value) ? null : value.Split("\n").First();
         }
 
         public string PrimaryContactPostcode { get; set; }
--- a/LBHTenancyAPI/Domain/TenancyListItem.cs
+++ b/LBHTenancyAPI/Domain/TenancyListItem.cs
@@ -5,12 +5,17 @@ namespace LBHTenancyAPI.Domain
 {
     public struct TenancyListItem
     {
-        public string TenancyRef { get; set; }
-        public string LastActionCode { get; set; }
-        public DateTime LastActionDate { get; set; }
+        public DateTime ArrearsAgreementStartDate { get; set; }
+
+        private string arrearsAgreementStatus;
+        public string ArrearsAgreementStatus
+        {
+            get => arrearsAgreementStatus;
+
+            set => arrearsAgreementStatus = value.Trim();
+        }
 
         private decimal currentBalance;
-
         public Decimal CurrentBalance
         {
             get => currentBalance;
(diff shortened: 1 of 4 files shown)


LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
```diff
@@ -19,28 +20,48 @@ public UhTenanciesGateway(string connectionString)
...
+                "ORDER BY arag.arag_startdate DESC, araction.action_date DESC"
+                ).ToList();
 
+            List<TenancyListItem> results = new List<TenancyListItem>();
+            foreach (var reference in tenancyRefs)
+            {
+                results.Add(all.First(e => e.TenancyRef == reference));
+            }
+
+            return results;
```
VladyslavAtamanyuk: Very nice refactor!

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -85,20 +86,19 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
...
+            Assert.Equal("ABC", receivedFirst.LastActionCode);
 
-            var receivedFirst = tenancies.Where(e => firstTenancy.TenancyRef == e.TenancyRef);
-            Assert.Equal(firstTenancyLatestActionDate, receivedFirst.First().LastActionDate);
-
-            var receivedSecond = tenancies.Where(e => secondTenancy.TenancyRef == e.TenancyRef);
-            Assert.Equal(secondTenancyLatestAgreementStartDate, receivedSecond.First().ArrearsAgreementStartDate);
+            var receivedSecond = tenancies.Find(e => e.TenancyRef == secondTenancy.TenancyRef);
+            Assert.Equal(secondTenancyLatestAgreementStartDate, receivedSecond.ArrearsAgreementStartDate);
+            Assert.Equal("Active    ", receivedSecond.ArrearsAgreementStatus);
```
VladyslavAtamanyuk: Why is there a bunch of spaces next to Active?

LBHTenancyAPI/Domain/TenancyListItem.cs
```diff
@@ -8,17 +8,27 @@ public struct TenancyListItem
         public string TenancyRef { get; set; }
         public string LastActionCode { get; set; }
         public DateTime LastActionDate { get; set; }
-        public double CurrentBalance { get; set; }
+
+        private decimal currentBalance;
```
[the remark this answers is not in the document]
StevenLeighton21: `CurrentBalance` is the public get/set, `currentBalance` is the private variable it's stored in. The reason for doing this is to override the public setter and ensure that the value plucked from the database is coerced into a Decimal of precision (0000000.00), I am unsure the rounding is correct (this _isn't_ bankers round I think)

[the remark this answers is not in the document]
StevenLeighton21: No need to apologise for asking questions, ever. If you don't understand what the code is doing, it may not be correct :)

LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
StevenLeighton21: Thanks. I feel this will undergo a further optimisation as plucking all results for this tenancy then plucking only the latest action and agreement is probably not performant, but could form a basis for the full Tenancy model...

LBHTenancyAPI/Domain/TenancyListItem.cs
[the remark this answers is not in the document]
VladyslavAtamanyuk: Ahhh i see thank you very much for the explanation!

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
StevenLeighton21: When you save a value to the server as CHAR(10), it pads it out to the size. I kind of like the fact that the test shows that I could insert something shorter and get out something padded to be the specified size.
At least, this tripped me up today and this is my best explanation, I wanted a test explicitly showing the weirdness because it forces you to dig deeper.

StevenLeighton21:
Have decided to trim char fields after discussing, see commit 434bf32

LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
zuren: Query's getting mad.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -78,27 +79,26 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
             TenancyListItem secondTenancy = InsertRandomisedTenancyListItem();
 
             DateTime firstTenancyLatestActionDate = firstTenancy.LastActionDate.AddDays(1);
-            InsertArrearsActions(firstTenancy.TenancyRef, "ABC", firstTenancyLatestActionDate);
+            InsertArrearsActions(firstTenancy.TenancyRef, "AB", firstTenancyLatestActionDate);
```
zuren: My problem with reusing tests for different behaviours like this, is that it's not obvious to me that the padding is under test. Having a specific test for that would act as documentation, which means it could be used as a specification for another implementation of ITenanciesGateway. whereas this is a side effect which may be ignored. Or introduce failures which aren't obvious fixes. e.g. in the event the trim code broke, would it be evident to a future developer that this test should remain the same, and that the trimming wasn't overlooked?

StubUniversalHousing/setup.sql
```diff
@@ -2,8 +2,8 @@
...
 USE StubUH;
 GO
-CREATE TABLE tenagree (tag_ref NVARCHAR(MAX), cur_bal FLOAT);
-CREATE TABLE araction (tag_ref NVARCHAR(MAX), action_code NVARCHAR(MAX), action_date SMALLDATETIME);
-CREATE TABLE arag (tag_ref NVARCHAR(MAX), arag_status NVARCHAR(MAX), start_date SMALLDATETIME);
-CREATE TABLE contacts (tag_ref NVARCHAR(MAX), con_name NVARCHAR(MAX), con_address NVARCHAR(MAX), con_postcode NVARCHAR(MAX), con_phone1 NVARCHAR(MAX));
+CREATE TABLE tenagree (tag_ref CHAR(11), cur_bal NUMERIC (9,3));
+CREATE TABLE araction (tag_ref CHAR(11), action_code CHAR(3), action_date SMALLDATETIME);
+CREATE TABLE arag (tag_ref CHAR(11), arag_status CHAR(10), arag_startdate SMALLDATETIME);
+CREATE TABLE contacts (tag_ref CHAR(11), con_name VARCHAR(73), con_address CHAR(200), con_postcode CHAR(10), con_phone1 CHAR(21));
```
zuren: These all came from real UH right?

StevenLeighton21: You were _there_

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -79,20 +81,65 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldReturnAllUnique
...
+            command.Parameters.Add("@tenancyRef", SqlDbType.Char);
+            command.Parameters["@tenancyRef"].Value = "not11chars";
+            command.Parameters.Add("@actionCode", SqlDbType.Char);
+            command.Parameters["@actionCode"].Value = "ee";
+            command.Parameters.Add("@aragStatus", SqlDbType.Char);
+            command.Parameters["@aragStatus"].Value = "status";
+            command.Parameters.Add("@postcode", SqlDbType.Char);
+            command.Parameters["@postcode"].Value = "pcode";
+            command.Parameters.Add("@phone", SqlDbType.Char);
+            command.Parameters["@phone"].Value = "phone";
```
zuren: Indentation off here.

PR #16, opened 2018-07-12
Add /tenancies endpoint and ListTenancies use case

zuren:
- Startup configures the dependency injection for the controller and the use case.
- The controller calls the ListTenancies use case.
- The ListTenancies use case uses it's injected ITenanciesGateway class to get tenancies by ref.

You can call the endpoint using:

```
curl http://localhost:32742/api/tenancies?tenancy_refs[0]=123&tenancy_refs[1]=456
```

### DIFF
--- a/LBHTenancyAPI/UseCases/ListTenancies.cs
+++ b/LBHTenancyAPI/UseCases/ListTenancies.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using LBHTenancyAPI.Gateways;
+
+namespace LBHTenancyAPI.UseCases
+{
+    public class ListTenancies
+    {
+        private ITenanciesGateway _tenanciesGateway;
+
+        public ListTenancies(ITenanciesGateway tenanciesGateway)
+        {
+            this._tenanciesGateway = tenanciesGateway;
+        }
+
+        public Response Execute(List<string> tenancyRefs)
+        {
+            var response = new Response();
+            var tenancies = _tenanciesGateway.GetTenanciesByRefs(tenancyRefs);
+
+            response.Tenancies = tenancies.ConvertAll<ResponseTenancy>(tenancy => new ResponseTenancy()
+                {
+                    TenancyRef = tenancy.TenancyRef,
+                    LastActionCode = tenancy.LastActionCode,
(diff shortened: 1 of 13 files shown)


LBHTenancyAPITest/Test/Controllers/TenanciesTest.cs
```diff
@@ -11,18 +13,198 @@ namespace LBHTenancyAPITest.Test.Controllers
...
+            {
+                Assert.Equal(expectedArguments, calledWith[0]);
+            }
+        }
+
+        private class ListTenanciesStub : IListTenancies
+        {
+            private readonly Dictionary<string, ListTenancies.ResponseTenancy> stubTenancies;
+
+            public ListTenanciesStub()
```
VladyslavAtamanyuk: Can we separate these private classes out into their own file?

LBHTenancyAPI/UseCases/ListTenancies.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            var response = new Response();
+            var tenancies = tenanciesGateway.GetTenanciesByRefs(tenancyRefs);
+
+            response.Tenancies = tenancies.ConvertAll(tenancy => new ResponseTenancy()
+                {
+                    TenancyRef = tenancy.TenancyRef,
+                    LastActionCode = tenancy.LastActionCode,
+                    LastActionDate = string.Format("{0:u}", tenancy.LastActionDate),
+                    CurrentBalance = tenancy.CurrentBalance.ToString(),
```
StevenLeighton21: Could we use `ToString("C")` to format as a currency? Or is this the responsibility of whatever will display the value? "C" is localised currency using banker's rounding IIRC

LBHTenancyAPITest/Test/Controllers/TenanciesTest.cs
zuren: Good idea, they may be reusable elsewhere. My only concern would be them becoming more generic to fit more test cases, whereas if they stay here they'll stay very specific to the tests I have. Do you have a strong preference? I could go either way honestly.

zuren:
Closed due to being merged into a separate PR, which has now merged into master.

PR #21, opened 2018-07-22
Get payments transactions

LBHRShetty:
- Added method GetPaymentTransactionDetails to TenanciesController.
- Added method signature GetPaymentTransactionsByTenancyRef to ITenanciesGateway.
- Added method GetPaymentTransactionsByTenancyRef to StubTenanciesGateway.
- Added method GetPaymentTransactionsByTenancyRef to UhTenanciesGateway class.
-Created test method WhenGivenATenancyRef_Index_ShouldRespondWithTenancyPaymentTransactionDetailsInfoForThatTenancy in TenanciesTest

### DIFF
--- a/.idea/.idea.LBHTenancyAPI/.idea/dataSources.xml
+++ b/.idea/.idea.LBHTenancyAPI/.idea/dataSources.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="DataSourceManagerImpl" format="xml" multifile-model="true">
+    <data-source source="LOCAL" name="@localhost" uuid="9ce64347-0068-4a08-8b84-1a0e1467c01c">
+      <driver-ref>sqlserver.ms</driver-ref>
+      <synchronize>true</synchronize>
+      <jdbc-driver>com.microsoft.sqlserver.jdbc.SQLServerDriver</jdbc-driver>
+      <jdbc-url>jdbc:sqlserver://localhost</jdbc-url>
+      <driver-properties>
+        <property name="applicationName" value="DataGrip" />
+      </driver-properties>
+    </data-source>
+  </component>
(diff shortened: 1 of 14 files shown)


.idea/.idea.LBHTenancyAPI/.idea/dataSources.xml
```diff
@@ -0,0 +1,14 @@
...
+      <driver-ref>sqlserver.ms</driver-ref>
+      <synchronize>true</synchronize>
+      <jdbc-driver>com.microsoft.sqlserver.jdbc.SQLServerDriver</jdbc-driver>
+      <jdbc-url>jdbc:sqlserver://localhost</jdbc-url>
+      <driver-properties>
+        <property name="applicationName" value="DataGrip" />
+      </driver-properties>
+    </data-source>
+  </component>
+</project>
```
zuren: Feel we should exclude editor specific files from git. I've been using a global gitignore to ignore .idea files in all projects, feel this is better as we don't have to include editor specific stuff in gitignores even.

See: https://help.github.com/articles/ignoring-files/ with `.idea` in the global gitignore file.

LBHTenancyAPI/Controllers/TenanciesController.cs
```diff
@@ -48,5 +48,56 @@ public TenanciesController(IListTenancies listTenancies)
...
+                        {"code_name", actionDiary.ActionCodeName},
+                        {"date", actionDiary.ActionDate.ToString()},
+                        {"comment", actionDiary.ActionComment}
+                    }
+                }
+            });
+
+            var result = new Dictionary<string, object>
+            {
+                {"arrearActionDiary", arrearActionDiary}
```
zuren: Should be standardising around snake_case instead of camelCase in JSON responses, so this should be `arrears_action_diary_events` or similar.

PR #23, opened 2018-07-24
Refactor

LBHRShetty:
Refactor the Payment Transactions and Arrears Action Diary Details for single tenancy Ref


StevenLeighton21:
Suggest we rebase the WIP commit

LBHTenancyAPI/Controllers/TenancyController.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using LBHTenancyAPI.UseCases;
+using Microsoft.AspNetCore.Mvc;
+
+namespace LBHTenancyAPI.Controllers
+{
+    [Produces("application/json")]
+    [Route("api/v1/tenancy")]
+    public class TenancyController : Controller
```
zuren: It's generally conventional in RESTful routing to group together individual entities with their collection in URLs. So typically you'd have a path like `/tenancies` as an endpoint for listing entities, and a path like `/tenancies/123` for a detailed view of the tenancy with id "123". Personally think we should follow this convention here, and move these methods into the TenanciesController.

LBHTenancyAPI/Controllers/TenancyController.cs
```diff
@@ -0,0 +1,72 @@
...
+        public async Task<IActionResult> PaymentTransactionDetails([FromQuery(Name = "tenancy_ref")] string tenancyRef)
+        {
+            var response = listAllPayments.Execute(tenancyRef);
+            var paymentsTransaction = response.PaymentTransactions.ConvertAll(paymentTrans => new Dictionary<string, object>
+            {
+                {"transaction_ref", paymentTrans.TransactionRef},
+                {"transaction_amount", paymentTrans.TransactionAmount},
+                {"transaction_date", paymentTrans.TransactionDate},
+                {"transaction_type", paymentTrans.TransactionType},
+                {"tenancy_ref", paymentTrans.TenancyRef},
```
zuren: I feel like it may not be relevant to include `tenancy_ref` in each transaction, as we'll know which tenancy it relates to already.

LBHTenancyAPI/Controllers/TenancyController.cs
```diff
@@ -0,0 +1,72 @@
...
+        [HttpGet]
+        public async Task<IActionResult> PaymentTransactionDetails([FromQuery(Name = "tenancy_ref")] string tenancyRef)
+        {
+            var response = listAllPayments.Execute(tenancyRef);
+            var paymentsTransaction = response.PaymentTransactions.ConvertAll(paymentTrans => new Dictionary<string, object>
+            {
+                {"transaction_ref", paymentTrans.TransactionRef},
+                {"transaction_amount", paymentTrans.TransactionAmount},
+                {"transaction_date", paymentTrans.TransactionDate},
+                {"transaction_type", paymentTrans.TransactionType},
```
zuren: I suggest we remove the "transaction_" prefix from these keys, as it's a bit redundant.

LBHTenancyAPI/Controllers/TenancyController.cs
```diff
@@ -0,0 +1,72 @@
...
+            var response = listAllArrearsActions.Execute(tenancyRef);
+            var arrearActionDiary = response.ActionDiaryEntries.ConvertAll(actionDiary => new Dictionary<string, object>
+            {
+                {"action_balance", actionDiary.ActionBalance},
+                {"action_code", actionDiary.ActionCode},
+                {"action_code_name", actionDiary.ActionCodeName},
+                {"action_date", actionDiary.ActionDate.ToString()},
+                {"action_comment", actionDiary.ActionComment},
+                {"universal_housing_username", actionDiary.UniversalHousingUsername},
+                {"tenancy_ref", actionDiary.TenancyRef}
```
zuren: Same here with the `tenancy_ref` key and the `action_` prefixes.

LBHTenancyAPI/Controllers/TenancyController.cs
StevenLeighton21: I would not have thought we would hit `tenancies/:Id/payments` for the payments for instance. I thought we would use `payments/:Id`.

Am I breaking convention too hard here?

LBHTenancyAPI/Controllers/TenancyController.cs
```diff
@@ -0,0 +1,72 @@
...
+            });
+
+            var result = new Dictionary<string, object>
+            {
+                {"arrears_action_diary", arrearActionDiary}
```
zuren: Suggest we call this `arrears_action_diary_events` or similar, as the "arrears action diary" is the name of the list, not the name of each item.

LBHTenancyAPI/Controllers/TenancyController.cs
zuren: In REST convention, those would mean different things.

- `/tenancies/:id/payments` would respond with payments for the given tenancy id.
- `/payments/:id` would return a single payment with the given payment id.

LBHTenancyAPI/Controllers/TenancyController.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public TenancyController(IListAllArrearsActions listAllArrearsActions)
+        {
+            this.listAllArrearsActions = listAllArrearsActions;
+        }
+
+        [Route("/payments")]
+        [HttpGet]
+        public async Task<IActionResult> PaymentTransactionDetails([FromQuery(Name = "tenancy_ref")] string tenancyRef)
```
zuren: Additionally, I think we'd have nicer URLs if we used path fragments instead of query params to pass in the tenancy ref. e.g. the full path would be `/tenancies/123/payments` instead of `/tenancies/payments?tenancy_ref=123`.

LBHTenancyAPI/Gateways/StubTenanciesGateway.cs
```diff
@@ -27,23 +28,33 @@ public List<TenancyListItem> GetTenanciesByRefs(List<string> tenancyRefs)
...
-            foreach (var actionDiary in tenancyRefs)
+
+            try
+            {
+                actionDiaryDetails.Add(StoredActionDiaryDetails[tenancyRef]);
+            }
+            catch (Exception)
             {
-                actionDiaryDetails.Add(StoredActionDiaryDetails[actionDiary]);
+                //do nothing
```
zuren: Feels dangerous to be catching all exceptions in these stub methods - what's raising what? Our stub shouldn't be brittle enough to have to do this IMO, and it feels like we could miss real issues with production code by doing this.

LBHTenancyAPI/UseCases/AllPaymentsForTenancy.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using LBHTenancyAPI.Gateways;
+
+namespace LBHTenancyAPI.UseCases
+{
+    public class AllPaymentsForTenancy : IListAllPayments
```
zuren: I think it's generally conventional to use the same name as the interface for the 'primary' implementation of something? So we'd either change the class to `ListAllPayments`, or the interface to `IAllPaymentsForTenancy`.

Thoughts?

LBHTenancyAPITest/Test/Controllers/TenancyTest.cs
```diff
@@ -0,0 +1,318 @@
...
+            var response = await GetPaymentTransactionDetails(allPayments, "NotHere");
+            Assert.NotNull(response);
+
+            var actualJson = ResponseJson(response);
+            var expectedJson = JsonConvert.SerializeObject
+            (
+                new Dictionary<string, object> {{"payment_transactions", new List<AllPaymentsForTenancy.PaymentTransaction>()}}
+            );
+
+            Assert.Equal(expectedJson, actualJson);
```
zuren: Wonder if this should actually be a 404 in this condition?

LBHTenancyAPITest/Test/Controllers/TenancyTest.cs
```diff
@@ -0,0 +1,318 @@
...
+            {
+
+                return new AllArrearsActionsForTenancy.ArrearsActionDiaryResponse
+                {
+                    ActionDiaryEntries = stubActionDiaryDetails.FindAll(e => e.TenancyRef == tenancyRef)
+                };
+            }
+        }
+    }
+}
```
zuren: Nice tests!

LBHTenancyAPI/Controllers/TenancyController.cs
StevenLeighton21: Understood, we'll get RESTy 👍

LBHTenancyAPI/Controllers/TenancyController.cs
LBHRShetty: done!

LBHTenancyAPI/Controllers/TenancyController.cs
LBHRShetty: done!

LBHTenancyAPI/Controllers/TenancyController.cs
LBHRShetty: done!

LBHTenancyAPI/Gateways/StubTenanciesGateway.cs
StevenLeighton21: Exceptions are flow management as well as error notification - in this case we're doing something that production is unable to do (pass null value to our internal struct in the use case) in order to test our _stubbed_ response. 
If we wanted to test what happens if you pass null to the endpoint (exception is raised), we'd test it elsewhere.

zuren: Don't fully follow. What exception exactly is raised here?

Could we at least specific the exception we expect to be raised, so it's not a catch all?

LBHTenancyAPITest/Test/Controllers/TenancyTest.cs
StevenLeighton21: Honestly, I wasn't sure. I'd be ok with 404!

LBHTenancyAPI/UseCases/AllPaymentsForTenancy.cs
zuren: There are - the only examples we have in our code of meeting an interface with a primary implementation follow that naming convention though.

ITenanciesGateway -> TenanciesGateway
IListTenancies -> ListTenancies

Totally open to changing it, just prefer consistency.

LBHTenancyAPI/Gateways/StubTenanciesGateway.cs
zuren: Ah, so why not:

```c#
if (StoredActionDiaryDetails.ContainsKey(actionDiary)) {
  actionDiaryDetails.Add(StoredActionDiaryDetails[actionDiary])
}
```

LBHTenancyAPI/UseCases/AllPaymentsForTenancy.cs
StevenLeighton21: A gateway interface is unlikely to be used by anything not a gateway or pretending to be one, so the same could be said for use cases I suppose? I Don't really mind either way personally

LBHTenancyAPI/Gateways/StubTenanciesGateway.cs
StevenLeighton21: I'm doing something along those lines as there's a refactor in the works to allow for removal of the tenancy ref anyway!

LBHTenancyAPI/Controllers/TenanciesController.cs
```diff
@@ -9,14 +9,18 @@
...
+        private readonly IListAllPayments listAllPayments;
+        private readonly IListAllArrearsActions listAllArrearsActions;
 
-        public TenanciesController(IListTenancies listTenancies)
+        public TenanciesController(IListTenancies listTenancies, IListAllArrearsActions listAllArrearsActions, IListAllPayments listAllPayments)
         {
             this.listTenancies = listTenancies;
+            this.listAllArrearsActions = listAllArrearsActions;
+            this.listAllPayments = listAllPayments;
         }
```
zuren: Bit sad that we need to construct all of these things in order to initialise the controller, when not all actions use all of these dependencies. Don't think it's a big deal, given .NET DI will do it for us, just probably makes testing more awkward.

Could maybe split it out so we have one action per controller at some point?

LBHTenancyAPI/Gateways/StubTenanciesGateway.cs
zuren: This isn't changing then?

LBHTenancyAPITest/Test/Controllers/GetAllPaymentsForTenancyRefTest.cs
```diff
@@ -0,0 +1,177 @@
...
+                    }
+
+                }
+            };
+
+            var actualResponse = ResponseJson(response);
+            var expectedJson = JsonConvert.SerializeObject(output);
+
+            Assert.Equal(expectedJson, actualResponse);
+        }
```
zuren: Missing triangulation here, meaning the controller could be responding with hardcoded data only suitable for our test.

LBHTenancyAPITest/Test/UseCases/ListAllPaymentsForTenancyTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            var payment = Fake.GeneratePaymentTransactionDetails();
+            gateway.SetPaymentTransactionDetails(payment.TenancyRef, payment);
+
+            var response = listAllPayments.Execute(payment.TenancyRef);
+
+            Assert.IsType(typeof(AllPaymentsForTenancy.PaymentTransactionResponse), response);
+        }
+
+        [Fact]
+        public void WhenATenancyRefIsGiven_ResponseShouldIncludePaymentsForThatTenancy_Example1()
```
zuren: Example1?

LBHTenancyAPI/Controllers/TenanciesController.cs
StevenLeighton21: I liked the idea of splitting into multiple controllers and maintaining RESTful routing, willing to put it down to tech debt at this point. I'm conscious we're doing some learning as we go (and at least in my case, I'm finding it helpful to keep it in one place)

LBHTenancyAPI/Gateways/StubTenanciesGateway.cs
StevenLeighton21: I don't know why the comment didn't collapse - it has changed

LBHTenancyAPITest/Test/Controllers/GetAllPaymentsForTenancyRefTest.cs
StevenLeighton21: Was brought up yesterday actually, I had come down on the side of - we're appealing to the gateway test to triangulate data and this test is checking json serialisation - the test isn't actually concerned about the content, only the json format. Thoughts?

I don't a problem with saying 'just triangulate anyway', but I chose the path of least resistance yesterday.

LBHTenancyAPI/Controllers/TenanciesController.cs
zuren: Fair enough, definitely fine to keep it that way if it's easier to learn! Don't feel there's any immediate need to change either,.

LBHTenancyAPITest/Test/Controllers/GetAllPaymentsForTenancyRefTest.cs
zuren: Sure, you're right in that we have separate tests for the use case and gateway which determine they respond correctly. We're only stubbing the use case here, so we don't even have a real need to respond with good test data so long as it meets the Response interface.

However, I feel this test would ideally verify that what the use case responds with is serialised properly into JSON, and at the moment it doesn't do so IMO. Two tests demonstrating different stub data being serialised would do so.

StevenLeighton21: Seems reasonable to me - I've updated with additional examples

LBHTenancyAPI/Controllers/TenanciesController.cs
StevenLeighton21: Bonus - if you look at the different tests, you don't actually have to pass in anything, you can pass null and it'll work until you try to reference it.

zuren: What test double is that? 🐺

LBHTenancyAPITest/Test/Controllers/GetAllPaymentsForTenancyRefTest.cs
zuren: Nice one.

PR #28, opened 2018-07-31
Fix issue where missing tenancy refs raise exception

StevenLeighton21:
Fixed issue where hitting the tenancies#index route raises an exception if you pass a list of refs where some (after the first) are invalid.

### DIFF
--- a/LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
+++ b/LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Data.SqlClient;
 using System.Linq;
 using Dapper;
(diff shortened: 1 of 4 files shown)


LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
```diff
@@ -55,7 +56,17 @@ public List<TenancyListItem> GetTenanciesByRefs(List<string> tenancyRefs)
...
+            {
+                try
+                {
+                    results.Add(all.First(e => e.TenancyRef == reference));
+                }
+                catch (InvalidOperationException)
+                {
+                    Console.Write($"No valid tenancy for ref: {reference}");
+                }
+            }
```
zuren: This is acceptable for now, but feel like we ought to be tracking both `results` and `errors`.

As an API consumer, I wouldn't want some tenancy refs to be apparently ignored without reason, it'd be a confusing experience.

Ideally I'd get back a result like:

```json
{
  "results": [
    { "tenancy_ref": "FAKE/01", ... }
  ],
  "errors": [
    { code: "MISSING_TENANCY", ref: "OTHER/01" }
  ]
```

StevenLeighton21: This would be way better, tech debt for now?

PR #29, opened 2018-07-31
Update endpoint to use the correct table values

StevenLeighton21:
I haven't updated to use the new standalone simulator - but I did copy the portion of the schema from it. This was slightly more expedient that swapping to the new simulator and retiring this project specific one, though I suggest doing so be made a priority to guarantee schema adherence.

### DIFF
--- a/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
+++ b/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
@@ -525,7 +525,7 @@ namespace LBHTenancyAPITest.Test.Gateways
                 items = new List<PaymentTransaction>();
                 string commandText =
 
-                    "INSERT INTO rtrans (tag_ref, trans_ref, prop_ref, trans_type, transaction_date, amount)" +
+                    "INSERT INTO rtrans (tag_ref, trans_ref, prop_ref, trans_type, post_date, real_value)" +
                     "VALUES (@tenancyRef, @transRef, @propRef, @transType, @transactionDate, @amount)";
 
                 foreach (int i in Enumerable.Range(0, num))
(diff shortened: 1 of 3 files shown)


kjdchapman:
Feels like needing to change the column names in multiple places (schema generation and gateway) is bad, but at least it's only two places. Guessing the Gateway does the job for anti-corruption for the most part.

LGTM

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -525,17 +525,17 @@ private List<PaymentTransaction> InsertRandomTransactions(string tenancyRef, int
                 items = new List<PaymentTransaction>();
                 string commandText =
 
-                    "INSERT INTO rtrans (tag_ref, trans_ref, prop_ref, trans_type, transaction_date, amount)" +
+                    "INSERT INTO rtrans (tag_ref, trans_ref, prop_ref, trans_type, post_date, real_value)" +
                     "VALUES (@tenancyRef, @transRef, @propRef, @transType, @transactionDate, @amount)";
 
                 foreach (int i in Enumerable.Range(0, num))
```
robinlacey: is there a reason you're using the Enumerable.Range rather than `for int i=0;i<num;i++`
it's a micro optimisation but foreach is a bit slower :)

craigjbass: I suspect the foreach with Enumerable.Range reads nicer to a Rubyist.

Isn't there a 

```c#
Enumerable.Range(0, num).Each((i) -> {

})
```

like syntax?

StevenLeighton21: Interesting thought - I actually just prefer the way foreach looks where the optimisation isn't necessary!

PR #32, opened 2018-08-02
Update endpoint to use correct table columns

StevenLeighton21:
- Switched made up column to type
- Corrected uh_username to username

**IMPORTANT**

We _must_ move to use the simulator project, but I did not migrate for this PR due to needing to migrate CI pipeline. It's not an insurmountable task, but more than I have time for right now.

### DIFF
--- a/LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
+++ b/LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
@@ -78,10 +78,10 @@ namespace LBHTenancyAPI.Gateways
                 "SELECT " +
                 "tag_ref as TenancyRef, " +
                 "action_code as ActionCode, " +
-                "action_code_name as ActionCodeName, " +
+                "action_type as Type, " +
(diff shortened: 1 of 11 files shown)


LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
```diff
@@ -78,10 +78,10 @@ public List<ArrearsActionDiaryEntry> GetActionDiaryEntriesbyTenancyRef(string te
                 "SELECT " +
                 "tag_ref as TenancyRef, " +
                 "action_code as ActionCode, " +
-                "action_code_name as ActionCodeName, " +
+                "action_type as Type, " +
```
LBHRShetty: Why did we change it to Type as in UH db it is code_name?

StevenLeighton21: `code_name` doesn't exist in the UH schema for the `araction` table.

PR #34, opened 2018-08-03
Fix tenancies route

StevenLeighton21:
- Route itself modified to be more RESTful
- Use case didn't pass through TenancyRef, so when the controller checked TenancyRef != null, it was always null. Resolved with test that shows it is passed through and encoded to JSON


StevenLeighton21:
I think we need a separate PR to add a proper unit test around the use case, this should/would have caught this I think.

PR #35, opened 2018-08-06
Single tenancy fix

StevenLeighton21:
Improved test coverage in an effort to diagnose issue causing empty staging responses. 
Endpoint works fine on local test:

### DIFF
--- a/LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
+++ b/LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
@@ -111,8 +111,8 @@ namespace LBHTenancyAPI.Gateways
         public Tenancy GetTenancyForRef(string tenancyRef)
         {
             var result = conn.Query<Tenancy>(
-                "SELECT DISTINCT" +
-                "(tenagree.tag_ref) as TenancyRef, " +
+                "SELECT DISTINCT " +
+                "tenagree.tag_ref as TenancyRef, " +
                 "tenagree.cur_bal as CurrentBalance, " +
                 "arag.arag_status as ArrearsAgreementStatus, " +
                 "arag.arag_startdate as ArrearsAgreementStartDate, " +
--- a/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
+++ b/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
@@ -207,7 +207,6 @@ namespace LBHTenancyAPITest.Test.Gateways
         {
             Tenancy expectedTenancy = CreateRandomSingleTenancyItem();
             InsertSingleTenancyAttributes(expectedTenancy);
-            expectedTenancy.ArrearsAgreements = InsertRandomAgreementDetails(expectedTenancy.TenancyRef, 6);
 
             var tenancy = GetSingleTenacyForRef(expectedTenancy.TenancyRef);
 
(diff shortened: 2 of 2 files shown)


LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
```diff
@@ -111,8 +111,8 @@ public List<PaymentTransaction> GetPaymentTransactionsByTenancyRef(string tenanc
         public Tenancy GetTenancyForRef(string tenancyRef)
         {
             var result = conn.Query<Tenancy>(
-                "SELECT DISTINCT" +
-                "(tenagree.tag_ref) as TenancyRef, " +
+                "SELECT DISTINCT " +
+                "tenagree.tag_ref as TenancyRef, " +
```
LBHRShetty: Why did you remove the brackets? I thought for distinct you have to put the brackets for the column you want to have distinct value

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -207,7 +207,6 @@ public void WhenGivenTenancyRef_GetSingleTenancyByRef_ShouldReturnTenancyWithBas
         {
             Tenancy expectedTenancy = CreateRandomSingleTenancyItem();
             InsertSingleTenancyAttributes(expectedTenancy);
-            expectedTenancy.ArrearsAgreements = InsertRandomAgreementDetails(expectedTenancy.TenancyRef, 6);
```
LBHRShetty: whitespace.

LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
StevenLeighton21: Good point - removed the `DISTINCT` since we're querying on a single reference and they will be unique, right?

Are we expecting tables that have duplicate tenancy references (if so, most of these joins will need reworking).

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
StevenLeighton21: Intentional :) I like to leave a line above and below the line that _executes_ the function I'm testing

StevenLeighton21: I removed the line adding 6 random details because it wasn't tested in the assertions in this particular test

LBHTenancyAPI/Gateways/UhTenanciesGateway.cs
```diff
@@ -105,14 +105,13 @@ public List<PaymentTransaction> GetPaymentTransactionsByTenancyRef(string tenanc
...
             ).ToList();
-
         }
 
         public Tenancy GetTenancyForRef(string tenancyRef)
         {
             var result = conn.Query<Tenancy>(
-                "SELECT DISTINCT" +
-                "(tenagree.tag_ref) as TenancyRef, " +
+                "SELECT " +
```
LBHRShetty: Why did we remove the brackets after distinct query.

StevenLeighton21: We won't need them at all for a general SELECT statement, since we're selecting everything we want specifically. We also shouldn't need DISTINCT in this case.

LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -230,6 +229,25 @@ public void WhenGivenTenancyRef_GetSingleTenancyByRef_ShouldReturnTenancyWithLat
...
+
+            expectedTenancy.ArrearsAgreements = InsertRandomAgreementDetails(expectedTenancy.TenancyRef, 6);
+
+            var tenancy = GetSingleTenacyForRef(expectedTenancy.TenancyRef);
+            Assert.Equal(expectedTenancy.PrimaryContactName, tenancy.PrimaryContactName);
+            Assert.Equal(expectedTenancy.PrimaryContactPostcode, tenancy.PrimaryContactPostcode);
+            Assert.Equal(expectedTenancy.PrimaryContactLongAddress, tenancy.PrimaryContactLongAddress);
+            Assert.Equal(expectedTenancy.PrimaryContactPhone, tenancy.PrimaryContactPhone);
+
+            Assert.Equal(5, tenancy.ArrearsAgreements.Count);
```
zuren: Why is this "5" when we appear to be including six random agreements?

StevenLeighton21: We're checking we get the latest five - I'm trusting that they are the latest and the 6th is the oldest based on the fact that that's done by the query ORDER BY

zuren: Should we be testing that order by somehow? i.e. creating six with escalating dates, and confirming we only get the latest five?

StevenLeighton21: I thought about it, I think we could test that the last item in the array is newer the the one left over. I will pick this up later and add the test cases.

PR #37, opened 2018-08-06
Added missing current balance in output json and TDDs

LBHRShetty:
Fixed the error of missing current balance in output json
Fixed missing current balance ibn example 1 and example 2 of tests

### DIFF
--- a/LBHTenancyAPI/Controllers/TenanciesController.cs
+++ b/LBHTenancyAPI/Controllers/TenanciesController.cs
@@ -136,6 +136,7 @@ namespace LBHTenancyAPI.Controllers
                     {
                         {"ref", tenancy.TenancyRef},
                         {"current_arrears_agreement_status", tenancy.ArrearsAgreementStatus},
+                        {"current_balance",tenancy.CurrentBalance},
                         {"primary_contact_name", tenancy.PrimaryContactName},
                         {"primary_contact_long_address", tenancy.PrimaryContactLongAddress},
                         {"primary_contact_postcode", tenancy.PrimaryContactPostcode}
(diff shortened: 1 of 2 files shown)


LBHTenancyAPITest/Test/Controllers/GetAllTenancyDetailsForGivenTenancyRefTest.cs
```diff
@@ -181,6 +181,7 @@ public async Task WhenGivenATenancyRef_TenancyDetail_ShouldRespondWithFormattedJ
             {
                 {"ref", "0test/01"},
                 {"current_arrears_agreement_status", "Breached"},
+                {"current_balance","23.01"},
```
StevenLeighton21: need spaces after commas! :D

LBHRShetty: fixed it now.

LBHTenancyAPI/Controllers/TenanciesController.cs
```diff
@@ -136,6 +136,7 @@ public async Task<IActionResult> GetTenancyDetails(string tenancyRef)
                     {
                         {"ref", tenancy.TenancyRef},
                         {"current_arrears_agreement_status", tenancy.ArrearsAgreementStatus},
+                        {"current_balance",tenancy.CurrentBalance},
```
LBHRShetty: fixed the space issue

StevenLeighton21:
I would consider renaming the PR to reference _tests_ not _TDDs_ but otherwise, LGTM (looks _great_ to me)

PR #38, opened 2018-08-07
Remove defensive programming for testing

StevenLeighton21:
We're seeing no data from this endpoint in live environments, so removing the checks in an effort to determine the cause.

### DIFF
--- a/LBHTenancyAPI/Controllers/TenanciesController.cs
+++ b/LBHTenancyAPI/Controllers/TenanciesController.cs
@@ -111,100 +111,66 @@ namespace LBHTenancyAPI.Controllers
         {
             Dictionary<string, object> result;
             var tenancyDetails = new Dictionary<string, object>();
-            try
+
+            var response = tenancyDetailsForRef.Execute(tenancyRef);
+
+            var tenancy = response.TenancyDetails;
+
+            if (tenancy.TenancyRef == null)
             {
-                if (string.IsNullOrWhiteSpace(tenancyRef))
+                result = new Dictionary<string, object>
                 {
-                    var errors = new List<APIErrorMessage>
-                    {
-                        new APIErrorMessage
-                        {
-                            developerMessage = "Invalid parameter - Tenancy Reference",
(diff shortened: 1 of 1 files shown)


LBHTenancyAPI/Controllers/TenanciesController.cs
```diff
@@ -111,101 +111,67 @@ public async Task<IActionResult> GetTenancyDetails(string tenancyRef)
...
-                    }
+                    {"tenancy_details", new Dictionary<string, object>()},
+                    {"latest_action_diary_events", new List<Dictionary<string, object>>()},
+                    {"latest_arrears_agreements", new List<Dictionary<string, object>>()}
                 };
-                var json = Json(errors);
-                json.StatusCode = 500;
-                return json;
+
+                return Ok(result);
```
zuren: Shouldn't be an `Ok` result here, right? That implies success 200.

LBHTenancyAPI/Controllers/TenanciesController.cs
```diff
@@ -111,101 +111,67 @@ public async Task<IActionResult> GetTenancyDetails(string tenancyRef)
...
+                    {"status", agreement.Status}
+                });
+
+            result = new Dictionary<string, object>
+            {
+                {"tenancy_details", tenancyDetails},
+                {"latest_action_diary_events", latestActionDiary},
+                {"latest_arrears_agreements", latestAgreement}
+            };
+
```
zuren: Lovely

LBHTenancyAPI/Controllers/TenanciesController.cs
StevenLeighton21: Well, the request was successful. There's just no matching data. I though that a valid but empty response should be OK, especially as the result classes as defined seem to only cover OK/NOK, though I'd definitely stand corrected

zuren: In this case I'd expect a 404, as the resource I was looking for was not found. As an API consumer, this lets me identify that my Tenancy ID is wrong.

PR #44, opened 2018-08-28
Address fix

StevenLeighton21:
- Move address and postcode from contacts to property where ref == tenagree.prop_ref
 - Also grabbing a few other necessary fields, tenure type, rent/service/other charges
- Both types of domain object have an object equality test, which should be sufficient to show that the new fields in the object returned by the gateway are the same as the randomly generated values inserted into the database.

### DIFF
--- a/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
+++ b/LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
@@ -116,10 +116,11 @@ namespace LBHTenancyAPITest.Test.Gateways
         public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldTrimCharacterFields()
         {
             string commandText =
-                "INSERT INTO tenagree (tag_ref) VALUES (@tenancyRef);" +
+                "INSERT INTO tenagree (tag_ref, prop_ref) VALUES (@tenancyRef, @propRef);" +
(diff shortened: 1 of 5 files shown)


LBHTenancyAPITest/Test/Gateways/UhTenanciesGatewayTest.cs
```diff
@@ -144,19 +149,29 @@ public void WhenGivenAListOfTenancyRefs_GetTenanciesByRefs_ShouldTrimCharacterFi
...
             Assert.Contains("phone                ", row.Values);
 
+            retrieved_values = db.Query("SELECT prop_ref, short_address, address1, post_code FROM property WHERE property.prop_ref = 'pref        '").ToList();
+            row = retrieved_values[0];
+
+            Assert.Contains("pref        ", row.Values);
+            Assert.Contains("pcode     ", row.Values);
+
+            row.TryGetValue("short_address", out var saved_address);
+            Assert.Equal(200, saved_address.ToString().Length);
```
zuren: Are we confident it won't be a 200 character string even if it's not assigned a value? i.e. the default value for this column isn't 200 chars of whitespace?

StevenLeighton21: I explicitly set it in the test, I'm not entirely sure we want a test around all nullable options in the database, though in this case a test to confirm a null value wouldn't break anything is probably useful. However, null values are handled by dapper by setting the field as null, and only not null values are trimmed.
The schema defines this address field as nullable with no default, and the other address field as nullable with a default empty space.

PR #48, opened 2018-09-03
Async awaited all the way through the stack as well as cleaning up a …

madetech-jeffpinkham:
Proposed new SQL query and a few performance improvements and slightly more idiomatic way of expressing serialisation.


StevenLeighton21:
Think we should close this and look at re-implementing?

madetech-jeffpinkham:
Yep

PR #53, opened 2018-09-12
this should fix the exception handler, we've got a custom middleware …

madetech-jeffpinkham:
Removed the default exception handler as we shouldn't need it


madetech-jeffpinkham:
cheers

PR #54, opened 2018-09-12
Feature/sentry exception logging

madetech-jeffpinkham:
Idiotmatic way of logging exceptions
Sentry exception logging by extending the default logging factory. 
By doing so anywhere that logs an exception using the ILogger<{class}> such as the CustomMiddlewareExceptionHandler will send it to sentry.


madetech-jeffpinkham:
That's a good suggestion, for the moment I'm inclined to leave it at as a verbose level as possible to help diagnose issues quickly, like we did earlier. Until the API and codebase stabilises a bit more, if that's ok :)

StevenLeighton21:
👍

PR #60, opened 2018-09-24
make API respond with API response object when it errors, instead of …

madetech-jeffpinkham:
API responds with standard "APIResponse" object when exceptions occur instead of just logging the error.


madetech-jeffpinkham:
Whenever an unexpected exception occurs error responses should look like:

```
{
  "data": null,
  "statusCode": 500,
  "error": {
    "IsValid": false,
    "Errors": [
      {
        "Message": "Value cannot be null.",
        "Code": null
      }
    ],
    "ValidationErrors": null
  }
}
```
So hopefully that will make it easier to consume in error scenarios. There is currently a mis-match however with the rest of the API endpoint as not all of them currently return this wrapping object so there is some work to get there

PR #62, opened 2018-09-26
Feature/get tenancy contacts

madetech-jeffpinkham:
Get Contacts from Dynamics 365.

1. Created API Exceptions
2. Used API Exceptions instead of ExecutionWrappers to simplify codebase.
3. Modified Exception Handling Middleware to handle Custom API Exceptions.
4. Created IHttpClient interface  to mock responses from Dynamics 365 API.
5. Created Dynamics 365 Client Factory to simplify talking to Dynamics 365.
6. Ported Dynamics 365 code from ManageATenancyAPI to generate AccessTokens and create clients.
7. Created Contacts Controller and RESTFul route ```
GET 
api/v1/tenancies/{TenancyAgreementReference}/contacts/```
8. Created contacts Domain object and separate mapping output object from the method to not confuse the 2 and create some seperation.
9. Hand mapped output object and didn't use AutoMapper because it's more performant.
10. Tests


LBH.Data.Domain/Contact.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using Newtonsoft.Json;
+
+namespace LBH.Data.Domain
+{
+
+    public class Contact
+    {
+        [JsonProperty("contactId")]
```
StevenLeighton21: Should be `contact_id` I think? For the other endpoints, we went with lower snake case field names, probably worth getting it consistent.

LBH.Data.Domain/Contact.cs
```diff
@@ -0,0 +1,56 @@
...
+        public Guid ContactId { get; set; }
+        [JsonProperty("emailAddress1")]
+        public string EmailAddress { get; set; }
+        [JsonProperty("hackney_uprn")]
+        public string UniquePropertyReferenceNumber { get; set; }
+        [JsonProperty("address1_line1")]
```
StevenLeighton21: Could we drop the `1` from the address fields? There's only one address per contact, no?

LBH.Data.Domain/Contact.cs
```diff
@@ -0,0 +1,56 @@
...
+        public string AddressLine2 { get; set; }
+        [JsonProperty("address1_line3")]
+        public string AddressLine3 { get; set; }
+        [JsonProperty("firstname")]
+        public string Firstname { get; set; }
+        [JsonProperty("lastname")]
+        public string LastName { get; set; }
+        [JsonProperty("fullname")]
+        public string FullName { get; set; }
+        [JsonProperty("hackney_larn")]
```
StevenLeighton21: Why do some fields have a `hackney_` prefix? I assume it means other potential contact sources could be supplying their own larn etc?

LBHTenancyAPI/Controllers/ContactController.cs
```diff
@@ -0,0 +1,33 @@
...
+    [Produces("application/json")]
+    [ProducesResponseType(typeof(APIResponse<object>), 400)]
+    [ProducesResponseType(typeof(APIResponse<object>), 500)]
+    public class ContactController : BaseController
+    {
+        private readonly IGetContactsForTenancyUseCase _createArrearsAgreementUseCase;
+
+        public ContactController(IGetContactsForTenancyUseCase createArrearsAgreementUseCase)
+        {
+            _createArrearsAgreementUseCase = createArrearsAgreementUseCase;
```
StevenLeighton21: Why does the contacts controller require a `createArrearsAgreementUseCase`?

LBH.Data.Domain/Contact.cs
```diff
@@ -0,0 +1,56 @@
+using System;
```
StevenLeighton21: I left a few comments on this file regarding naming of the fields, appreciating that the 'outward' facing naming is handled elsewhere, this means naming here for consistency with dynamics etc. is fine and treat my other comments as suggestions!

LBHTenancyAPI/UseCases/Contacts/Models/ContactsForTenancy.cs
```diff
@@ -0,0 +1,84 @@
...
+            Title = contact.Title;
+            FullAddressDisplay = contact.FullAddressDisplay;
+            FullAddressSearch = contact.FullAddressSearch;
+            PostCode = contact.PostCode;
+            DateOfBirth = contact.DateOfBirth;
+            HackneyHomesId = contact.HackneyHomesId;
+            Age = contact.Age;
+        }
+
+        [JsonProperty("contactId")]
```
StevenLeighton21: Need to decide if this should be lower snake case like our other JSON output, or if this output is more comfortable for other API consumers (and this endpoint is useful outside of our project'

LBHTenancyAPI/Controllers/ContactController.cs
madetech-jeffpinkham: copy paste error

LBH.Data.Domain/Contact.cs
madetech-jeffpinkham: thats just what the fields are called in dynamics 365 contact table

LBHTenancyAPI/Gateways/Contacts/Dynamics365RestApiContactsGateway.cs
```diff
@@ -0,0 +1,89 @@
...
+            var response = await httpClient.GetAsync(query, cancellationToken).ConfigureAwait(false);
+            if(!response.IsSuccessStatusCode)
+                throw new Dynamics365RestApiException(response);
+
+            //call dynamics 365
+            var json = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+            //maybe no records found so return null
+            if (string.IsNullOrEmpty(json))
+                return null;
+            var contacts = JsonConvert.DeserializeObject<GetContactsResults>(json.Replace("contact1_x002e_", ""));
```
LBHRShetty: Why are we doing this json.Replace("contact1_x002e_"?

LBH.Data.Domain/Contact.cs
madetech-jeffpinkham: It more denotes Line number

LBHTenancyAPI/Gateways/Contacts/Dynamics365RestApiContactsGateway.cs
madetech-jeffpinkham: Currently the json properties that gets returned from the fetch look like ```'contact1_x002e_telephone3'``` 
```'x002e'```
 is due to unicode encoding and the property being linked from the contact table.
It should actually look like 
```'contact1_._telephone3'```
There are 2 approaches to this:
Set an alias on the fetch query or strip the prefix.
On this occasion I've stripped the prefix because I wanted to preserve the raw names from the Contact table, and have the Contact Domain object map 1 to 1 with that.

File: LBH.Data.Domain/Contact.cs (hunk 1 above)
Comment (madetech-jeffpinkham) [reply LBH.Data.Domain/Contact.cs]: Yep good point

File: LBHTenancyAPI/Infrastructure/Dynamics365/Exceptions/Dynamics365RestApiException.cs [hunk 8]
```diff
@@ -0,0 +1,18 @@
...
+using LBHTenancyAPI.Infrastructure.UseCase.Execution;
+
+namespace LBHTenancyAPI.Infrastructure.Dynamics365.Exceptions
+{
+    public class Dynamics365RestApiException : APIException
+    {
+        public ExecutionError ExecutionError { get; set; }
+
+        public Dynamics365RestApiException(HttpResponseMessage response)
+        {
```
LBHRShetty: Will it be useful to have specific error message thrown before we get CRM access token as discussed earlier

LBHTenancyAPI/Settings/CRM/Dynamics365Settings.cs
```diff
@@ -0,0 +1,11 @@
+namespace LBHTenancyAPI.Settings.CRM
+{
+    public class Dynamics365Settings
+    {
+        public string OrganizationUrl { get; set; }
+        public string ClientId { get; set; }
```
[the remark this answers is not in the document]
madetech-jeffpinkham: Should be ClientId and TenantID should be TenantId

LBH.Data.Domain/Contact.cs
StevenLeighton21: I meant more `address1_line1` could be `address_line1` and so on

LBH.Data.Domain/Contact.cs
StevenLeighton21: In this particular file, I'd stick with that. You already dropped it from the model so, 👍!

LBHTenancyAPI/Gateways/Contacts/Dynamics365RestApiContactsGateway.cs
StevenLeighton21: Good candidate for moving to a named function or a constant IMO, since that would clarify it. You could have `const UNWANTED_ENCODED_PREFIX = 'contact1_x002e';` for instance and it would make it more obvious to anyone maintaining.

LBH.Data.Domain/Contact.cs
madetech-jeffpinkham: The default in .Net is CamelCasing, there will be some work to get it all snake casing, I'll have to swap out the default Json Contract serialiser.

LBH.Data.Domain/Contact.cs
madetech-jeffpinkham: That's how it gets returned from Dynamics365, we can create alias' in the fetchXML but for the moment I wanted to leave it mapped 1 to 1 in case we wanted to do CRUD operations later on.

LBHTenancyAPI/Gateways/Contacts/Dynamics365RestApiContactsGateway.cs
madetech-jeffpinkham: Yeah good suggestion 👍

LBHTenancyAPI/UseCases/Contacts/Models/ContactsForTenancy.cs
madetech-jeffpinkham: This is a wider issue we need to tackle as it causes regression issues with the create action diary event endpoint. I can set the default json serialiser to be snake casing for everything.

StevenLeighton21: The other endpoints use snake case? Or am I just misremembering? I think we already have a mix at this point, and we could certainly make it consistent but I'd consider that a separate task

LBHTenancyAPI/Settings/CRM/Dynamics365Settings.cs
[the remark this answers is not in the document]
StevenLeighton21: I believe this is the preferred way but actually goes against https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71). Specifically, capitalising two character acronyms. I would argue that Id should be treated the same as Url personally :) (so no change)

LBH.Data.Domain/Contact.cs
madetech-jeffpinkham: cool thanks :)

LBHTenancyAPI/UseCases/Contacts/Models/ContactsForTenancy.cs
madetech-jeffpinkham: Ok, yep so do you want snake casing on the returned contacts?

madetech-jeffpinkham: snake casing on outputted contacts list

PR #65, opened 2018-10-01
Bug/sql connection

madetech-jeffpinkham:
Fix sql connection not getting closed/disposed by using "using" which auto disposes of any IDisposable once execution leaves the "using" scope.


madetech-jeffpinkham:
Correct

madetech-jeffpinkham:
Hmm might be a different kind of issue.

PR #67, opened 2018-10-05
Feature/simple search

madetech-jeffpinkham:
Simple Search on 5 criteria
First Name
Last Name
Post Code
Contains address
Tag reference


StevenLeighton21:
Suggest using `git rebase -i` to squash commit history. IIRC it doesn't like to squash commits where you merged master into this branch, so just go back to the last merge commit.

madetech-jeffpinkham:
> Suggest using `git rebase -i` to squash commit history. IIRC it doesn't like to squash commits where you merged master into this branch, so just go back to the last merge commit.

Is that the only command I need to use to get the desired result? I don't ever rebase really so I want to make sure I get it right.

LBHTenancyAPI/Gateways/Search/SearchGateway.cs
```diff
@@ -0,0 +1,94 @@
...
+                    end
+                    if(@page > 0)
+                    begin
+                    Set @Lower = (@pageSize * @page) + 1
+                    Set @Upper = @Lower + @pageSize -1
+                    end
+
+                    SELECT
+                    Seq,
+                    ArrearsAgreementStartDate,
```
StevenLeighton21: Do we _need_ agreement status for search results?

LBHTenancyAPI/Gateways/Search/SearchGateway.cs
```diff
@@ -0,0 +1,94 @@
...
+                        ON property.prop_ref = tenagree.prop_ref
+                        RIGHt JOIN  arag AS arag WITH(NOLOCK)
+                        ON arag.tag_ref = tenagree.tag_ref
+                        WHERE tenagree.tag_ref = @searchTerm
+                        OR member.forename = @searchTerm
+                        OR member.surname = @searchTerm
+                        OR property.short_address like '%'+ @searchTerm +'%'
+                        OR property.post_code like  '%'+ @searchTerm +'%'
+                    )
+                    t
```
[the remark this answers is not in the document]
madetech-jeffpinkham: An alias for the results from the query in the '( )' SQL seems to complain if you don't type a name in there. 't' is not a good name. Any suggestions? :)

LBHTenancyAPI/Gateways/Search/SearchGateway.cs
madetech-jeffpinkham: That's a good question, I talked about trying to make the response as similar to the list page, however there has been some divergence because we are not getting last action code and last action date. Let me ask the team if they think it should be there.

LBHTenancyAPITest/EF/Entities/ArrearsAgreement.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+using Dapper;
+
+namespace LBHTenancyAPITest.EF.Entities
```
StevenLeighton21: Still using EF in this revision?

madetech-jeffpinkham: Nope, I however did re-use the entities, as the Domain is different, I'll adjust.

LBHTenancyAPITest/Helpers/Fake.cs
```diff
@@ -159,5 +161,96 @@ public static ArrearsActionResponse CreateArrearsActionAsync(ArrearsActionCreate
...
+                    .RuleFor(property => property.title, (fake, model) => "Mr")
+                    .RuleFor(property => property.age, (fake, model) => fake.Random.Int(20, 50))
+                    .RuleFor(property => property.responsible, (fake, model) => true)
+                    .RuleFor(property => property.person_no, (fake, model) => fake.IndexFaker)
+                    ;
+                var member = faker.Generate();
+                return member;
+            }
+
+            public static TenancyAgreement GenerateFakeTenancy()
```
StevenLeighton21: Is this in order to generate fake EF objects? I'm unclear if this is staying/necessary in this revision. That said, couldn't you just `return faker.Generate();`?

StubUniversalHousing/setup.sql
```diff
@@ -4,15 +4,277 @@ GO
 USE StubUH;
```
StevenLeighton21: I think we should remove every field not explicitly tested from this file. We intentionally only added the minimum necessary to test our code to avoid leaking more of the contents of the schema than necessary.

StevenLeighton21:
> > Suggest using `git rebase -i` to squash commit history. IIRC it doesn't like to squash commits where you merged master into this branch, so just go back to the last merge commit.
> 
> Is that the only command I need to use to get the desired result? I don't ever rebase really so I want to make sure I get it right.

It opens an interactive rebase commit in the default editor. I use https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History as a reference, and go for `f, fixup` instead of `s, squash` to keep code but discard commit messages in most cases.

StevenLeighton21:
I don't think I have much more to add in terms of review, I can't see any major holes but it's a large PR. I'll give some time for additional reviewers, but its a +1 from me.

StubUniversalHousing/setup.sql
madetech-jeffpinkham: done

madetech-jeffpinkham: done

LBHTenancyAPI/Gateways/Search/SearchGateway.cs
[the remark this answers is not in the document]
madetech-jeffpinkham: Changed 't' to 'orderByWithSequenceSubQuery' which is a bit more descriptive.

LBHTenancyAPITest/Helpers/Fake.cs
madetech-jeffpinkham: I wanted a way to interact with the Database involving POCO classes, and initially was using EF, but with the removal of EF I still wanted a way to interact with the Database. I was originally using a generic generate but fields were being truncated and so causing errors. So I went in and specified the maxlength of the generated fields so as to not cause errors.

madetech-jeffpinkham:
Ok thanks

LBHTenancyAPI/Gateways/Search/SearchGateway.cs
madetech-jeffpinkham: I've asked the question, and will await a reply.

madetech-jeffpinkham:
@StevenLeighton21 I'm going to merge this in, I've been trying to squash the changes but I need a bit more practise with that.

PR #68, opened 2018-10-08
Missing to lower on the fields

madetech-jeffpinkham:
For some reason the to lower casing was missing


StevenLeighton21:
Undo, undo! Better fix the checks....

madetech-jeffpinkham:
Haha too late now!! Yes I'm just looking into it the error is odd. Fear not I won't merge until everything is ok.

PR #69, opened 2018-10-08
Set currency to GBP

madetech-jeffpinkham:
Fix currency and set to GBP. Staging is currently returning '¤' whatever that is. Setting to GBP '£'


StevenLeighton21:
Note super confident about this. It will almost guaranteed break the front end app, to to brittleness on that side.
The "c" setting made all the currency values come through in JSON as "¤12,345", in the rush to get around this and get data on the page, we're just deleting them (turns out, we actually need to convert back to numbers in some cases, which is an entirely different kettle of dangerous, dangerous fish). Changing these to '£' would be much more meaningful, but will break everything. Removing the currency symbols entirely _shouldn't_.

madetech-jeffpinkham:
Ahh ok maybe we should not return it as a currency and just a decimal/float.

PR #70, opened 2018-10-09
Bug/simple search currency

madetech-jeffpinkham:
Remove the currency icon as the Front End is stripping it anyway.
Set the current balance to be a floating point number in Json (decimal) in C#.
Add currency Symbol Field if the Front End wishes to use it.
Return en-gb culture currency Symbol.


StevenLeighton21:
Is it possible to run this and the front end locally, point both to staging and generate a screenshot of a detail page (list page will require the rails API) to prove it does not require a PR on the front end side?

madetech-jeffpinkham:
@StevenLeighton21 it's a new endpoint and should require a PR to implement the search functionality. It shouldn't however affect any existing functionality.

madetech-jeffpinkham:
@StevenLeighton21 I would need to be walked through doing that with the Front End as I've not pulled it down before.

PR #74, opened 2018-10-11
Bug/search returns results with no tenancy agreement

madetech-jeffpinkham:
Fixed a few things:
1) Filter out anyone not attached to a tenancy agreement.
2) Some search results for tenants who appear to have outstanding balances are not attached to properties and weren't being returned.
3) Sort by member.surname ,member.forename as requested. 782ms - 1200ms on dev
4) Suggest asking for indexing to be done on those columns


mrosel:
I don't see any issues with the code. Looks solid. 
However, I don't feel we have adequate tests around 
https://github.com/LBHackney-IT/LBHTenancyAPI/pull/74/files#diff-803db12498fd58a3c4f50377b3230f19R71
since we should have an exact match on tenancy_ref

madetech-jeffpinkham:
@mrosel well spotted 👍 I've added additional.

PR #75, opened 2018-10-12
explicitly close connections

madetech-jeffpinkham:
explicitly close connections as wrapping in a "using" block appears to not close them


madetech-jeffpinkham:
Yep agreed.

PR #76, opened 2018-10-15
Feature/health check

madetech-jeffpinkham:
Sql Health Check 

Can add other health checks and returns composite health check result

```
{
  "checkStatus": 2,
  "description": "SqlConnectionHealthCheck: Successfully retrieved 1 record from database - 46ms",
  "data": {
    "SqlConnectionHealthCheck": {}
  },
  "results": {
    "SqlConnectionHealthCheck": {
      "checkStatus": 2,
      "data": {},
      "description": "Successfully retrieved 1 record from database - 46ms"
    }
  }
}
```

### DIFF
--- a/LBHTenancyAPI/Infrastructure/Health/SqlConnectionHealthCheck.cs
+++ b/LBHTenancyAPI/Infrastructure/Health/SqlConnectionHealthCheck.cs
@@ -1,12 +1,9 @@
-using System;
-using System.Collections.Generic;
-using System.Collections.ObjectModel;
-using System.Data.SqlClient;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Dapper;
+using LBHTenancyAPI.Factories;
 using Microsoft.Extensions.HealthChecks;
 using Microsoft.Extensions.Logging;
 
@@ -14,12 +11,12 @@ namespace LBHTenancyAPI.Infrastructure.Health
 {
     public class SqlConnectionHealthCheck:IHealthCheck
     {
-        private readonly string _connectionString;
+        private readonly ISqlConnectionFactory _sqlConnectionFactory;
         private readonly ILogger<SqlConnectionHealthCheck> _logger;
 
-        public SqlConnectionHealthCheck(string connectionString, ILogger<SqlConnectionHealthCheck> logger)
+        public SqlConnectionHealthCheck(ISqlConnectionFactory sqlConnectionFactory, ILogger<SqlConnectionHealthCheck> logger)
         {
-            _connectionString = connectionString;
+            _sqlConnectionFactory = sqlConnectionFactory;
             _logger = logger;
         }
         public async ValueTask<IHealthCheckResult> CheckAsync(CancellationToken cancellationToken = new CancellationToken())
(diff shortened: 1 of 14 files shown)


LBHTenancyAPI/Program.cs
```diff
@@ -13,6 +15,7 @@ public static void Main(string[] args)
         public static IWebHost BuildWebHost(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
                 .UseStartup<Startup>()
+                .UseKestrel(server=> server.Limits.KeepAliveTimeout = TimeSpan.FromMinutes(30))
```
StevenLeighton21: What's the significance of this?

madetech-jeffpinkham: Server stays alive for 30 minutes after the last request, the current setting seems to be less than this.

LBHTenancyAPI/appsettings.json
```diff
@@ -3,16 +3,18 @@
     "IncludeScopes": false,
     "Debug": {
       "LogLevel": {
-        "Default": "Information",
-        "System": "Information",
-        "Microsoft": "Information"
+        "Default": "Debug",
```
StevenLeighton21: Intentionally reverted back to debug? It didn't seem helpful, just spammy.

madetech-jeffpinkham: It's helpful to me, when trying to debug issues. It's spammy but also goes into the level of detail I need to diagnose issues remotely.

madetech-jeffpinkham: The reason it seems spammy was because it goes into the level of detail of thread execution, which is helpful as there is threading in the API. At this point I expect the issues to be subtle ones and not obvious ones.

StevenLeighton21: Fair enough :)

LBHTenancyAPI/Program.cs
StevenLeighton21: Does it make a difference? Feels like a change that shouldn't be buried in this PR as it'd be hard to track without using `blame`

madetech-jeffpinkham: Good point

madetech-jeffpinkham: Removed

LBHTenancyAPI/Infrastructure/Health/SqlConnectionHealthCheck.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Dapper;
+using Microsoft.Extensions.HealthChecks;
+using Microsoft.Extensions.Logging;
+
+namespace LBHTenancyAPI.Infrastructure.Health
+{
+    public class SqlConnectionHealthCheck:IHealthCheck
```
StevenLeighton21: I'm in two minds about testing this class. I don't see any tests around it, but I'm not sure adding them really tells too much other than the sql connection works, which is covered elsewhere.

LBHTenancyAPITest/Test/Controllers/HealthcheckTest.cs
```diff
@@ -1,32 +1,62 @@
...
+using System.Threading.Tasks;
 using LBHTenancyAPI.Controllers;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.HealthChecks;
+using Moq;
 using Xunit;
 
 namespace LBHTenancyAPITest.Test.Controllers
 {
     public class HealthcheckTest
```
StevenLeighton21: I don't see a test mocking out the case where the health check fails to connect to the sql server. Again, I'm not sure how much overhead of testing I'd insist on for a health check but in the absence of a unit test around the connection class, I'd suggest adding one.

LBHTenancyAPI/Infrastructure/Health/SqlConnectionHealthCheck.cs
madetech-jeffpinkham: Done some light testing around this, healthy and unhealthy scenario.

PR #85, opened 2018-10-23
Feature/search multi faceted search

madetech-jeffpinkham:
Multi Faceted Search
Versioning
Swagger documentation versioning
1 failing test, need someone to review.


LBHTenancyAPI/Extensions/String/StringExtensions.cs
```diff
@@ -0,0 +1,15 @@
+namespace LBHTenancyAPI.Extensions.String
+{
+    public static class StringExtensions
+    {
+        public static bool IsNullOrEmptyOrWhiteSpace(this string str)
+        {
+            return string.IsNullOrEmpty(str) || string.IsNullOrWhiteSpace(str);
+        }
+
+        public static bool IsNotNullOrEmptyOrWhiteSpace(this string str)
```
StevenLeighton21: Is this extension useful compared to `! IsNullOrEmptyOrWhiteSpace`

madetech-jeffpinkham: Yep saves me having to ! IsNullOrEmptyOrWhiteSpace every time. It's used like 10+ times in searchgateway v2

LBHTenancyAPITest/Test/Controllers/V2/SearchControllerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+        private Mock<ISearchTenancyUseCase> _mock;
+
+        public SearchControllerTests()
+        {
+            _mock = new Mock<ISearchTenancyUseCase>();
+            _classUnderTest = new SearchController(_mock.Object);
+        }
+
+        [Fact]
+        public async Task GivenValidSearchTenancyRequest_WhenCallingGet_THenShouldReturn200()
```
StevenLeighton21: `GivenValidSearchTenancyRequest_WhenCallingGet_THenShouldReturn200` should be `GivenValidSearchTenancyRequest_WhenCallingGet_ThenShouldReturn200`

StevenLeighton21:
There's a _lot_ in here that is just V1 - V2 auto refactor changes. Would be great in future to split this out into a commit and have a commit for the files that need to be reviewed. Also, the test files are so large the diffs aren't loaded by default, which makes it hard to review. I don't think there's any value in trying to change it now, but it could help smoother reviews in future

madetech-jeffpinkham:
Yeah agreed I thought that simply versioning it would make it difficult to review. As I'm moving files around.

StevenLeighton21:
I don't think there's much you can do about it due to the nature of the change, but we _could_ do things like PR the version switch, and PR the new endpoint separately (even in stages, PR gateway & test first, etc)

madetech-jeffpinkham:
Yep, I agree I should have probably asked for a review sooner.

StevenLeighton21:
Seems great to me but there is a lot in there, so would like to allow time for other reviews before merging.

PR #88, opened 2018-10-26
Start Class Level documentation - Pull XML comments into swagger docs

madetech-jeffpinkham:
Pull XML comments into Swagger Docs 
screen shot 2018-10-26 at 13 24 34

### DIFF
--- a/LBHTenancyAPI/Factories/ISqlConnectionFactory.cs
+++ b/LBHTenancyAPI/Factories/ISqlConnectionFactory.cs
@@ -8,7 +8,7 @@ namespace LBHTenancyAPI.Factories
     public interface ISqlConnectionFactory
     {
         /// <summary>
-        /// Creates a SQL connection however doesn't open it
+        /// Creates an instance of a SQL connection object without opening a connection
         /// </summary>
         /// <returns></returns>
         SqlConnection Create();
--- a/LBHTenancyAPI/Factories/ISqlConnectionFactory.cs
+++ b/LBHTenancyAPI/Factories/ISqlConnectionFactory.cs
@@ -1,9 +1,16 @@
-﻿using System.Data.SqlClient;
+using System.Data.SqlClient;
 
 namespace LBHTenancyAPI.Factories
 {
+    /// <summary>
+    /// Factory for creating SQL connections 
+    /// </summary>
     public interface ISqlConnectionFactory
     {
+        /// <summary>
+        /// Creates a SQL connection however doesn't open it
+        /// </summary>
+        /// <returns></returns>
         SqlConnection Create();
     }
 }
--- a/LBHTenancyAPI/Factories/SqlConnectionFactory.cs
+++ b/LBHTenancyAPI/Factories/SqlConnectionFactory.cs
@@ -3,6 +3,9 @@ using Microsoft.Extensions.Logging;
 
 namespace LBHTenancyAPI.Factories
 {
+    /// <summary>
+    /// Simple Factory for creating SQL connections 
+    /// </summary>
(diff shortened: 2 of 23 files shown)


LBHTenancyAPI/Factories/SqlConnectionFactory.cs
```diff
@@ -15,7 +18,7 @@ public SqlConnectionFactory(string connectionString, ILogger<SqlConnectionFactor
         }
 
         /// <summary>
-        /// Creates sql connection
+        /// Creates sql connection however doesn't open it
```
StevenLeighton21: Suggest wording along the lines of `Creates SQL connection object without opening a connection`

LBHTenancyAPI/Factories/ISqlConnectionFactory.cs
```diff
@@ -1,9 +1,16 @@
...
 
 namespace LBHTenancyAPI.Factories
 {
+    /// <summary>
+    /// Factory for creating SQL connections 
+    /// </summary>
     public interface ISqlConnectionFactory
     {
+        /// <summary>
+        /// Creates a SQL connection however doesn't open it
```
StevenLeighton21: Suggest avoiding however as described in other comments

LBHTenancyAPI/Factories/WCFClientFactory.cs
```diff
@@ -2,8 +2,17 @@
 
 namespace LBHTenancyAPI.Factories
 {
+    /// <summary>
+    /// Simple Factory for WCF clients
+    /// </summary>
     public class WCFClientFactory : IWCFClientFactory
     {
+        /// <summary>
+        /// Creates WCF Clients however doesn't open connection
```
StevenLeighton21: Suggest avoiding however as described in other comments

LBHTenancyAPI/Gateways/V1/Search/SearchGateway.cs
```diff
@@ -7,7 +7,11 @@
 using LBHTenancyAPI.UseCases.V1.Search.Models;
 
 namespace LBHTenancyAPI.Gateways.V1.Search
-{
+{
+    /// <summary>
+    /// Search Gateway V1 Searches for tenants attached to tenancies
+    /// Currently connects to UH Database via SQL connection and SQL queries
```
StevenLeighton21: Not sure about use of currently. Is it expected this will change and the documentation will need maintenance? If so, suggest just not mentioning it, rather than relying on future maintainers to update docs for accuracy. 
If it is unlikely to change, suggest rewording .

LBHTenancyAPI/UseCases/V1/Search/SearchTenancyUseCase.cs
```diff
@@ -6,33 +6,44 @@
...
+        //Dependency Injected Search Gateway
         private readonly ISearchGateway _searchGateway;
 
         public SearchTenancyUseCase(ISearchGateway searchGateway)
         {
             _searchGateway = searchGateway;
         }
 
+        /// <summary>
+        /// Execute the
```
StevenLeighton21: The suspense is killing me

LBHTenancyAPI/Gateways/V1/Search/SearchGateway.cs
StevenLeighton21: If you feel it's worth keeping the line, just drop 'Currently'? My comment is more around the assumption of future maintenance of the docs, and the wording implies it will be necessary! If you feel that the nature of the connection will change in the near future, I'd consider not documenting it, because I feel the most likely scenario is that document is not maintained and becomes misleading.

PR #94, opened 2018-10-30
set default for value

madetech-jeffpinkham:
set default value so you don't always have to pass it in


LBHTenancyAPI/Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /src
 COPY LBHTenancyAPI.sln ./
 COPY LBHTenancyAPI/LBHTenancyAPI.csproj LBHTenancyAPI/
 
-ARG VERSION_SUFFIX
+ARG VERSION_SUFFIX=0
```
cob16: Can I suggest that it be something like "NO VERSION SUFFIX SUPPLYED"

madetech-jeffpinkham: Not sure how I accomplish this I'm afraid.

PR #110, opened 2019-01-29
Add PaymentRef and StartDate fields to Get tenancy route

alexdesi:
- Fix property ref always being null


LBHTenancyAPI/Gateways/V1/UhTenanciesGateway.cs
```diff
@@ -25,43 +25,45 @@ public List<TenancyListItem> GetTenanciesByRefs(List<string> tenancyRefs)
...
-                    tenagree.cur_bal as CurrentBalance, 
-                    tenagree.prop_ref as PropertyRef, 
-                    tenagree.tenure as Tenure, 
-                    tenagree.rent as Rent, 
-                    tenagree.service as Service, 
-                    tenagree.other_charge as OtherCharge, 
+                  @"SELECT
+                    tenagree.tag_ref as TenancyRef,
+                    tenagree.u_saff_rentacc as PaymentRef,
+                    tenagree.cot as StartDate,
```
mrosel: nice find!

LBHTenancyAPI/UseCases/V1/TenancyDetailsForRef.cs
```diff
@@ -20,28 +21,16 @@ public TenancyResponse Execute(string tenancyRef)
...
-                    Rent = tenancyResponse.Rent.ToString("C"),
-                    Service = tenancyResponse.Service.ToString("C"),
-                    OtherCharge = tenancyResponse.OtherCharge.ToString("C"),
-                    PrimaryContactName = tenancyResponse.PrimaryContactName,
-                    PrimaryContactLongAddress = tenancyResponse.PrimaryContactLongAddress,
-                    PrimaryContactPostcode = tenancyResponse.PrimaryContactPostcode,
-                    ArrearsActionDiary = new List<ArrearsActionDiaryEntry>(),
-                    ArrearsAgreements = new List<ArrearsAgreement>()
-                };
+                response.TenancyDetails = new Tenancy();
```
mrosel: Would it be too complicated at this point to return a 404 instead?

Makefile
```diff
@@ -13,7 +13,7 @@ setup: docker-build
 
 .PHONY: serve
 serve:
-	docker run -p 3000:80 -it --rm lbhtenancyapi
+	docker-compose run -p 3000:80 -e UH_URL=$(TEST_UH_CONNECTION_STRING) --rm lbhtenancyapi
```
mrosel: :tada: one step closer to real dev/test env's

LBHTenancyAPI/UseCases/V1/TenancyDetailsForRef.cs
cob16: The change is larger than just the use case. I stated it before and it got out of hand.

PR #113, opened 2019-07-09
Set isCommentOnly true on all action diary requests

cob16:


LBHTenancyAPI/Gateways/V2/Arrears/UniversalHousing/Impl/ArrearsServiceRequestBuilder.cs
```diff
@@ -33,6 +33,7 @@ public ArrearsActionCreateRequest BuildNewActionDiaryRequest(ActionDiaryRequest
                     ActionCategory = request.ActionCategory,
                     ActionCode = request.ActionCode,
                     Comment = request.Comment,
+                    IsCommentOnly = true,
```
LBHRShetty: is always mandatory to keep as true?

cob16: The current understanding is that this keeps us from resetting state in UHW

gw-hackney: @LBHRShetty for MA yes, it will always be the case. However, we're interested in this case to see if it makes any difference anyway.

LBHRShetty: ok great thanks :)

PR #115, opened 2019-09-23
returning joint tenants as one record

elena-vi:


LBHTenancyAPITest/Test/Gateways/V1/UhTenanciesGatewayTest.cs
```diff
@@ -376,6 +376,23 @@ public void WhenGivenTenancyRef_GetSingleTenancyByRef_ShouldReturnTenancyWithBas
             Assert.Equal(expectedTenancy.CurrentBalance, tenancy.CurrentBalance);
         }
 
+        [Fact]
+        public void WhenGivenJointTenancyRef_GetSingleTenancyByRef_ShouldReturnTenancyWithBasicDetails()
```
georgeschena: That method name is Looooooong

georgeschena:
Can I have a description of the PR please

LBHTenancyAPITest/Test/Gateways/V1/UhTenanciesGatewayTest.cs
elena-vi: that's a test name

kjdchapman:
Is there any performance implication in getting many records instead of the first? I remember us making some scary noises around UH performance :grimacing:

elena-vi:
> Is there any performance implication in getting many records instead of the first? I remember us making some scary noises around UH performance 😬

tbh, I'm not sure. Hoping that for a single tenancy ref we'll get either 1 or 2 records.
I kinda want to plug it into staging and see how it handles it.

georgeschena:
If it won't have much performance impact then Im happy to approve.

PR #116, opened 2019-09-25
refactor the way joint tenancies are merged

elena-vi:
Initially staging was falling over when trying to search by the address `sara lane court`, I thought that it might have been because the way we were joining joint tenancies was slow.

After refactoring performance has increased from 3394 ms to 3337 ms (faster by 57ms) also now the code responsible for that looks cleaner

### DIFF
--- a/LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
+++ b/LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
@@ -91,7 +91,7 @@ namespace LBHTenancyAPI.UseCases.V2.Search
             {
                 foreach (var duplicateTenancy in duplicateTenancies)
                 {
-                    var jointTenancy = duplicateTenancy[0];
+                    var jointTenancy = duplicateTenancy.FirstOrDefault();
                     foreach (var tenancy in duplicateTenancy.Where(thing =>
                         jointTenancy.PrimaryContactName != thing.PrimaryContactName))
                     {
--- a/LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
+++ b/LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
@@ -36,23 +36,17 @@ namespace LBHTenancyAPI.UseCases.V2.Search
             if (response == null)
                 return new SearchTenancyResponse();
 
-            var allTenancyListItems = response.Results;
+            var allTenancyListItems = response.Results.ToList();
 
-            var uniqueTenancyListItems = allTenancyListItems.GroupBy(t => t.TenancyRef)
-                .Where(g => g.Count() == 1)
-                .SelectMany(g => g).ToList();
-
-            var groupsOfDuplicatesTenancyListItems =
-                allTenancyListItems.GroupBy(t => t.TenancyRef)
-                .Where(g => g.Count() > 1);
-
-
-            updateUniqueListItems(uniqueTenancyListItems,groupsOfDuplicatesTenancyListItems);
+            if (allTenancyListItems.Count > 1)
+            {
+                allTenancyListItems = MergeJointTenancies(allTenancyListItems);
+            }
 
             //Create real response
             var useCaseResponse = new SearchTenancyResponse
             {
-                Tenancies = uniqueTenancyListItems.ConvertAll(tenancy => new SearchTenancySummary
+                Tenancies = allTenancyListItems.ConvertAll(tenancy => new SearchTenancySummary
                 {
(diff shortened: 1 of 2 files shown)


LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
```diff
@@ -72,22 +66,49 @@ public async Task<SearchTenancyResponse> ExecuteAsync(SearchTenancyRequest reque
...
+                    duplicateTenancies.Add(new List<TenancyListItem> {tenancy, duplicateTenancy});
+                    newResult.Remove(duplicateTenancy);
+                }
+                else
+                {
+                    newResult.Add(tenancy);
+                }
+            }
+
+            if (duplicateTenancies.Count <= 0) return newResult;
```
kjdchapman: Why don't you have an `Empty()`, C# :(

We could use this here:
```suggestion
            if (!duplicateTenancies.Any()) return newResult;
```

But that `!` is fairly quiet, so you may not like this more.

LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
```diff
@@ -72,22 +66,49 @@ public async Task<SearchTenancyResponse> ExecuteAsync(SearchTenancyRequest reque
...
             {
-                var jointTenancies = grouping.ToList();
-                var jointTenancy = jointTenancies[0];
-                jointTenancies.ForEach(delegate(TenancyListItem dup)
+                foreach (var duplicateTenancy in duplicateTenancies)
                 {
-                    if (jointTenancy.PrimaryContactName != dup.PrimaryContactName)
+                    var jointTenancy = duplicateTenancy[0];
```
kjdchapman: I'd use a method call here, but you could argue either is clearer/terser.
```suggestion
                    var jointTenancy = duplicateTenancy.FirstOrDefault();
```

LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
```diff
@@ -72,22 +66,49 @@ public async Task<SearchTenancyResponse> ExecuteAsync(SearchTenancyRequest reque
...
+                    foreach (var tenancy in duplicateTenancy.Where(thing =>
```
kjdchapman: I can't find a better word than thing, but I definitely got confused by what it was briefly!

LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
```diff
@@ -72,22 +66,49 @@ public async Task<SearchTenancyResponse> ExecuteAsync(SearchTenancyRequest reque
...
-                        jointTenancy.PrimaryContactName += $" & {dup.PrimaryContactName}";
+                        jointTenancy.PrimaryContactName += $" & {tenancy.PrimaryContactName}";
                     }
-                });
-                uniqueTenancyListItems.Add(jointTenancy);
+
+                    var alreadyPresent = newResult.Find(
+                        x => x.TenancyRef == jointTenancy.TenancyRef
+                    );
+                    if (alreadyPresent.TenancyRef == null)
```
kjdchapman: Same comment as [line 79](https://github.com/LBHackney-IT/LBHTenancyAPI/pull/116/files#diff-bb6c80b29592da59501bd4120b95c146R79) - having to check if `tenancyref` is null, rather than the tenancy itself, makes this harder to understand.

That might be necessary because of the shape of the data?

LBHTenancyAPI/UseCases/V2/Search/SearchTenancyUseCase.cs
```diff
@@ -72,22 +66,49 @@ public async Task<SearchTenancyResponse> ExecuteAsync(SearchTenancyRequest reque
...
-        private void updateUniqueListItems(List<TenancyListItem> uniqueTenancyListItems,
-            IEnumerable<IGrouping<string, TenancyListItem>> groupsOfDuplicatesTenancyListItems)
+        private static List<TenancyListItem> MergeJointTenancies(List<TenancyListItem> allTenancyListItems)
         {
-            foreach (var grouping in groupsOfDuplicatesTenancyListItems)
+            var duplicateTenancies = new List<List<TenancyListItem>>();
+            var newResult = new List<TenancyListItem>();
+            foreach (var tenancy in allTenancyListItems)
+            {
+                var duplicateTenancy = allTenancyListItems.Find(
```
StevenLeighton21: Suggest using `Where` over `Find` if you're using conditionals to check (it'll be faster)

PR #117, opened 2019-10-04
Get num_bedrooms from property table in UH

georgeschena:
WHAT: Get num_bedrooms column from UH

WHY: So we can send this to the frontend and show the user the number of bedrooms for a tenancy.

### DIFF
--- a/LBHTenancyAPITest/Test/UseCases/V1/TenancyDetailsForRefTest.cs
+++ b/LBHTenancyAPITest/Test/UseCases/V1/TenancyDetailsForRefTest.cs
@@ -114,7 +114,6 @@ namespace LBHTenancyAPITest.Test.UseCases.V1
 
             tenancy.NumberOfBedrooms = null;
 
-
             gateway.SetTenancyDetails(tenancy.TenancyRef, tenancy);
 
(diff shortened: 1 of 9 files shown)


LBHTenancyAPITest/Test/UseCases/V1/TenancyDetailsForRefTest.cs
```diff
@@ -103,6 +105,22 @@ public void WhenATenancyRefIsGiven_ResponseShouldIncludeTenancyDetails()
...
+
+        [Fact]
+        public void WhenATenancyRefIsGiven_AndTheNumberOfBedroomsIsNull()
+        {
+            var gateway = new StubTenanciesGateway();
+            var tenancy = Fake.GenerateTenancyDetails();
+
+            tenancy.NumberOfBedrooms = null;
+
+            gateway.SetTenancyDetails(tenancy.TenancyRef, tenancy);
```
kjdchapman: This is a surprising method signature, and it wasn't initially clear what the test is doing.

Any worries about this test suite? I haven't had the chance to look at it yet. Not saying we need to change anything now.

PR #118, opened 2019-10-07
only liable(responsible) tenants are shown as joint tenants

elena-vi:
users need to know if the tenancy is a joint tenancy with all names on the tenancy so I can ensure that case details are disclosed to the correct person. Meaning the person(s) marked liable

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=MA&modal=detail&selectedIssue=MA-69

### DIFF
--- a/LBHTenancyAPI/Gateways/V1/UhTenanciesGateway.cs
+++ b/LBHTenancyAPI/Gateways/V1/UhTenanciesGateway.cs
@@ -171,7 +171,7 @@ namespace LBHTenancyAPI.Gateways.V1
                     LEFT JOIN property WITH(NOLOCK)
                     ON property.prop_ref = tenagree.prop_ref
                     Left JOIN dbo.member member WITH(NOLOCK)
-                    ON member.house_ref = tenagree.house_ref
+                    ON member.house_ref = tenagree.house_ref and member.responsible = 1
                     WHERE tenagree.tag_ref = @tRef
                     ORDER BY arag.arag_startdate DESC",
                     new {tRef = tenancyRef.Replace("%2F", "/")}
--- a/LBHTenancyAPI/Gateways/V1/UhTenanciesGateway.cs
+++ b/LBHTenancyAPI/Gateways/V1/UhTenanciesGateway.cs
@@ -171,8 +171,8 @@ namespace LBHTenancyAPI.Gateways.V1
                     LEFT JOIN property WITH(NOLOCK)
                     ON property.prop_ref = tenagree.prop_ref
                     Left JOIN dbo.member member WITH(NOLOCK)
-                    ON member.house_ref = tenagree.house_ref and member.responsible = 1
-                    WHERE tenagree.tag_ref = @tRef
+                    ON member.house_ref = tenagree.house_ref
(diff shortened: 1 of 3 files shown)


LBHTenancyAPI/Gateways/V1/UhTenanciesGateway.cs
```diff
@@ -171,7 +171,7 @@ public Tenancy GetTenancyForRef(string tenancyRef)
                     LEFT JOIN property WITH(NOLOCK)
                     ON property.prop_ref = tenagree.prop_ref
                     Left JOIN dbo.member member WITH(NOLOCK)
-                    ON member.house_ref = tenagree.house_ref
+                    ON member.house_ref = tenagree.house_ref and member.responsible = 1
```
bjpirt: wouldn't this AND clause go in the WHERE conditions?

elena-vi: yes

LBHTenancyAPITest/Test/Gateways/V1/UhTenanciesGatewayTest.cs
```diff
@@ -122,10 +123,21 @@ private Tenancy GenerateTenancy(string tenency_ref = "")
             expectedTenancy.prop_ref = expectedProperty.prop_ref;
             expectedTenancy.start_date = expectedTenancy.start_date;
             TestDataHelper.InsertTenancy(expectedTenancy, _databaseFixture.Db);
-            //member 1
+
             var expectedMember = Fake.UniversalHousing.GenerateFakeMember();
-            expectedMember.house_ref = expectedTenancy.house_ref;
-            TestDataHelper.InsertMember(expectedMember, _databaseFixture.Db);
+            if (household != null && household.Any())
```
kjdchapman: Would it be better to default the household to an empty array and make this check simpler? Or do we expect the caller to pass `null` to us explicitly sometimes?

LBHTenancyAPITest/Test/Gateways/V1/UhTenanciesGatewayTest.cs
```diff
@@ -379,11 +391,20 @@ public void WhenGivenTenancyRef_GetSingleTenancyByRef_ShouldReturnTenancyWithBas
...
+            var expectedTenant2 = Fake.UniversalHousing.GenerateFakeMember(responsible: true);
+
+            var householdMembers = new List<Member>()
+            {
+                expectedTenant, expectedTenant2,
+                Fake.UniversalHousing.GenerateFakeMember(responsible: false),
+                Fake.UniversalHousing.GenerateFakeMember(responsible: false)
+            };
+
+            Tenancy expectedTenancy = GenerateTenancy(household: householdMembers);
```
kjdchapman: Could we pull up the context of this test into the naming? I had to read it a couple of times to "get" it was four tenants, two of which are 'responsible', and only the responsible ones show up.

`public void GivenMixedTenants_GetSingleTenancyByRef_ReturnsOnlyResponsibleTenants()`?

Still a bit of a monster! Not sure how to do this without context blocks, not familiar with the language/framework.

PR #120, opened 2019-10-09
Add num_bedroms to Tenancies controller

georgeschena:
Added num_bedroms to the tenancies controller


LBH.Data.Domain/Tenancy.cs
```diff
@@ -108,6 +108,7 @@ public string PaymentRef
 
         public DateTime? StartDate { get; set; }
 
+        private string numberOfBedrooms;
```
kjdchapman: I'm not sure I understand this change. What is this private field doing?

Is this related to another change we made? I can't see anything in #117 that helps me either :confused:

georgeschena: We wanted to be consistent with how the other fields are declared within this file

georgeschena: The reason we weren't pulling in the number of bedrooms is because we forgot to add it into the controller

kjdchapman: Worried about us doing things for consistency without knowing what they're doing. 

My (limited) understanding of C# suggests this backing field isn't needed - do the tests pass without it?

PR #122, opened 2019-11-18
Allow API to be ran as part of a service

richrace:


bjpirt:
For context - this relates to https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/242 and https://github.com/LBHackney-IT/lbh-income-api/pull/252 which will allow us to spin up the Manage Arrears front end along will all associated service dependencies

PR #126, opened 2020-05-19
Remove myself from active maintainers

cob16:


cob16:
@elena-vi I don't have merge rights would you please do the honours

PR #134, opened 2021-07-22
Query performance fix

LBHSPreston:
## What
- Added optimisation to a list of string to improve dapper query performance.
- Increased container memory to help improve performance.

## Why
- The get transactions API call was taking extremely long to complete making the application almost unusable.  It is expected that adding this conversion would improve performance.

## Link to task
https://app.clickup.com/t/pdejuf


felipenudge (commented):
Review done

